PR #12, opened 2017-03-11
melee attack-range leeway mechanic

gemt:
In reference to: https://elysium-project.org/bugtracker/issue/2307

It should work for player vs player and player vs npc. I'm not sure if this mechanic should work for npcs vs players as well (e.g if a mob is chasing a player), so I ignored that.

### DIFF
--- a/src/game/Objects/Object.h
+++ b/src/game/Objects/Object.h
@@ -172,7 +172,7 @@ enum MovementFlags
     MOVEFLAG_HOVER              = 0x40000000,
     MOVEFLAG_INTERNAL           = 0x80000000,
 
-    // Nostalrius : ne peut etre present avec MOVEFLAG_ROOT (freez client sinon)
+    // Nostalrius : Can not be present with MOVEFLAG_ROOT (otherwise client freeze)
     MOVEFLAG_MASK_MOVING        =
         MOVEFLAG_FORWARD | MOVEFLAG_BACKWARD | MOVEFLAG_STRAFE_LEFT | MOVEFLAG_STRAFE_RIGHT |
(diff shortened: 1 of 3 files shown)


gemt:
On land, this seems to work in accordance with how the client indicates in range/out of range. However, when in water, judging from the client, there is no leeway at all.

gemt:
There are some more comments in https://elysium-project.org/bugtracker/issue/2307 now. People seems to think the mechanic should take effect for NPC vs player too.

In addition there is some code here: 
https://github.com/elysium-project/server/blob/master/src/game/Spells/Spell.cpp#L6126
that seems to be an attempt at a leeway mechanic, but the solution is wrong as far as I can understand.
The solution already in place on elysium adds 1.5yd if target is moving, and 1.5yd if player is moving. If both are moving that adds up to 3yd total. There is no requirement that both players move, and that they move at least at 70% of normal runspeed.

It also only works for melee spells (abilities), and not autoattack. I am currently assuming it should work for autoattacks as well.

gemt:
commit 22397c0 should be close to a complete solution for melee leeway solution.

However, there was an old implementation in Spells.cpp#L6126 that applied a somewhat bugged leeway range (wrong ranges and missing conditions) higher on the callstack. I removed it, and I don't know if it can have any side-effects. It needs some testing.

This solution still only works for player vs player or player vs npc. Not npc vs player or npc vs npc. That can easily be changed by disabling an if-statement.

sidsukana:
Thank you, we will test it and tell about results.

sidsukana:
I hope you will support your fix if suddenly it will be necessary :)

gemt:
Sure thing.

robinsch:
I will just leave this here as a ref:
http://paste2.org/2FUCZ4Ix
https://github.com/TrinityCore/TrinityCore/commit/14b93c04ee07e9fe6e2767460e7fc7cc2b3e6808

PR #13, opened 2017-03-11
Fix for sweeping strike with execute

Tenarie:
The current sweeping strike just makes an AA.

https://elysium-project.org/bugtracker/issue/604


sidsukana:
What about Execute? There is no check of 20% HP.

Tenarie:
Looking into it :), thought we were going for 1.9 patch notes.
But i've read that spells and talents should follow 1.12 : ).

Tenarie:
Ay, i'd like to hear what you think.
I think it's pretty spot on to the real vanilla.

m-laniakea:
Reopen

PR #17, opened 2017-03-12
Fix for ritual part of RFD escort.

brotalnia:
Currently on Elysium there are some bugs in the ritual script for the quest Extinguishing the Idol in RFD. Belnistrasz chases mobs while he is supposed to be channeling his ritual spell, and when the event is complete he doesn't spawn the gameobject "Belnistrasz's Brazier", so you can't turn in the quest. This commit resolves these bugs.


src/scripts/kalimdor/the_barrens/razorfen_downs/razorfen_downs.cpp
```diff
@@ -272,6 +272,7 @@ struct npc_belnistraszAI : public npc_escortAI
                 switch (m_uiRitualPhase)
                 {
                     case 0:
+                        DoCast(m_creature, 23973, true); // roots self so he stays put while channeling
```
Zerixx: Enum or define. Magic numbers can get annoying to deal with.

PR #18, opened 2017-03-13
Fix: Level up hp/mana text

EinBaum:
As described here:

https://elysium-project.org/bugtracker/issue/3263


EinBaum:
Edited: Previous commit included int in the mana amount (mistake).

PR #19, opened 2017-03-13
Bug 3688. Change damage school from frost to nature on Noxxion and his adds

henkethebenke:
Fixed [a bug from the bugtracker](https://elysium-project.org/bugtracker/issue/3688). Noxxion and his adds were dealing frost damage instead of nature.


StadenElysium:
Seems reasonable enough to me - I can't find any evidence that to the contrary.

PR #20, opened 2017-03-13
.gocorpse command

EinBaum:
* Behaves like .goname
* Works with offline players
* Teleports to own corpse or your target's corpse if you don't specify a name
* NOTE: Dungeon / BG corpses will not be visible to you if you are not in the player's group


src/game/Commands/Level1.cpp
```diff
@@ -588,6 +588,31 @@ bool ChatHandler::HandleGonameCommand(char* args)
...
+
+    Corpse* corpse = sObjectAccessor.GetCorpseForPlayerGUID(target_guid);
+    if (!corpse)
+    {
+        PSendSysMessage(LANG_COMMAND_TELE_NOTFOUND);
+        SetSentErrorMessage(true);
+        return false;
+    }
+
+    float x, y, z;
```
StadenElysium: Can you declare the variables on the same line as their first use?
e.g. `float x = func();` rather than `float x; x = func();`

EinBaum: Sure, fixed.

PR #28, opened 2017-03-14
Update mob_generic_creature.cpp

shmgiy429:
Non-elite creatures should cast a spell 20% of the time not 80%?


Tenarie:
Yea i've felt this was wrong aswell. They are really spamming abilities.

Tenarie:
But you'll most likely have to find proof. Maybe go ahead and find someone farming in Tyrs hand back in vanilla ?

StadenElysium:
Code already works as intended.

PR #37, opened 2017-03-14
fix for issue 34 - ah-humanoid spells

Fr4nsson:
Many of the npc's have their stats and spells wrong. Fix down below.


shmgiy429:
So you've entered new values in for these NPC's but, all we can see from these changes are the current values you want to Insert; is there a way to see what was incorrectly in the table for comparison?

PR #43, opened 2017-03-15
[Core/Scripts]: Implement Event Based Timers

Zerixx:
* Event based timers. Allows us to use scheduled events rather than the
time diff checking.

* Easier to write, and easier to read for newbies.

* Simplifies code flow and looks far more professional to the trained
eye.

* Comes from TrinityCore so we can cross reference scripting techniques
and start coding in the 21st century on this core..

* As a proof of concept, I have rewritten Lucifron (first boss in MC) to
show an example of how this might look in the future with the other
scripts.

Edit: TC Sources https://github.com/TrinityCore/TrinityCore/commit/7ea8bad07f64ed5886130fcc8016c0c72a32758e


Inifield:
I see i'm not the only one who backported this from TC into mangos :p

Zerixx:
@Inifield Much better way of writing bosses, imo.

PR #50, opened 2017-03-16
Command: Add .groupinfo command

StadenElysium:


gearselysium:
Server crash when used

PR #52, opened 2017-03-17
Simplifying if statement

henkethebenke:
https://www.viva64.com/en/b/0476/#ID0EERBI


m-laniakea:
Remove the unnecessary parentheses, and you're good.

PR #56, opened 2017-03-18
Trade window glitch: The scam prevention system prevented players from accepting

FractalCabush:
Within 3 seconds of opening the trade window it is not possible to accept a trade. The scam prevention part of the trade system was not set to look for the next modification timestamp and thus only did its job on the initial first 3 seconds. I implemented it according to what i think the original developer was after. Added: Scam prevention system with variable delay length, no more glitchy trade windows in in-game trade windows.


EinBaum:
Watch out, you have mixed tabs with spaces.

FractalCabush:
@EinBaum you are right, i will fix that :)

e92b:
if this works, gj :)

PR #58, opened 2017-03-19
Potential fix for Application freeze

elysium-duks:
Potential fix for kernal based lockup of Mangos.

PR #61, opened 2017-03-19
EventProcessor::GetEvents should return const reference instead of value

gemt:
In reference to comment on issue #58 
I don't know this piece of code, so someone should check if the fix can cause other issues, but it does prevent the server from crashing when player changes map.


Zerixx:
LGTM. This will fix the crash.

PR #66, opened 2017-03-19
Make general channels multi-language

Neo2003:
French, German, Spanish, English,... clients should be able to see what other write in general channels like Trade, Guild Recruitment, Local Defense...

Also in Defense channels, user rank should be displayed.

Basically, general channels are decoded using client locale and special names are used internally. When a message must be sent to the channel, it's decoded pet client locale to sent with the proper channel name for their localization.

Name decoding is based on AreaTable.dbc, so server must have all languages dbc to support them like this:
dbc/AreaTable.dbc <- the English one
dbc/frFR/AreaTable.dbc <- French one
dbc/deDE/AreaTable.dbc <- German one
...
You probably have them all already.

I could not use the AreaTable you moved to database since it's not localized.

Neo2003
Signed-off-by: Neo2003 <Neo.2003@Hotmail.fr>

### DIFF
--- a/src/game/Chat/Channel.cpp
+++ b/src/game/Chat/Channel.cpp
@@ -25,12 +25,12 @@
 #include "SocialMgr.h"
 #include "MasterPlayer.h"
 
-Channel::Channel(const std::string& name)
+Channel::Channel(const std::string& name, const std::string& originalName, LocaleConstant locale)
     : m_announce(true), m_moderate(false), m_name(name), m_flags(0), m_channelId(0),
         m_securityLevel(0), m_area_dependant(true), m_levelRestricted(true)
 {
     // set special flags if built-in channel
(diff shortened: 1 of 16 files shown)


Neo2003:
Hi

Please note that this feature was developed for Vengeance WoW 1.12.1 one year ago.
Thanks for Laizerox research around Honor ranks and defense channel restriction and for testing this with me, and Cyberium who spent time porting this to 2.4.3.

Neo2003

StadenElysium:
PR doesn't compile, reverted.

Inifield:
just fix the compile it's not that hard...

StadenElysium:
It will be fixed when I have the time to fix it.

Neo2003:
It's now fixed, I did remove honor code that GCC does not like on Linux.
Sorry, I did make it on Windows with VC14 without testing on Linux, my bad before making a pool request..

PR #69, opened 2017-03-19
.lookup account iponline

EinBaum:
Suggested by GM Skeith.

Same as .lookup account ip but shows only online accounts.


elysium-duks:
Code reviewed, looks great. - Passed off to Shenna for acceptance.

PR #79, opened 2017-03-20
Make general channels multi-language v2

Neo2003:
Now compile properly on Windows AND Linux.
I dropped honor part for now so that gcc is happy.

Neo2003

### DIFF
--- a/src/game/Chat/Channel.cpp
+++ b/src/game/Chat/Channel.cpp
@@ -625,11 +625,7 @@ void Channel::Say(ObjectGuid p, const char *what, const char* chanName, uint32 l
     int32 playerRank = 0;
 
     if (plr)
-    {
         sec = plr->GetSession()->GetSecurity();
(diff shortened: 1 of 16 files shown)


Neo2003:
I have Defense channel requirements coded and working on both Windows and Linux, but since it depends on this one, I will open a PR for it when this one is merged.

m-laniakea:
The general channels were changed specifically not to be multi-language.
We want English to be spoken in general, major cities, etc.

People are free to communicate in whatever language they wish in private or in guild chat.

Neo2003:
I understand, but from non English client, we can't talk nor read what is written in general channels because the channels have not the same name. The purpose of this patch is to enable from any language client to read/talk in these channels. After, for me it's a question of server rules to enforce the use of English only on general channels which seams obvious.

StadenElysium:
Reverted due to major performance issues that require investigation.

ryancheung:
😅

Issue #71, opened 2017-03-19
Druid aquatic form missing


SinisterMatti:
On Anathema since the last patch when shifting into aquatic form druids stay in the caster form, but still have the swim speed buff and underwater breathing.

henkethebenke:
Are you alliance or horde? Isn't [this row](https://github.com/elysium-project/server/blob/43452c0d9cdce841bef71c592a381c38fb68d37a/src/game/Spells/SpellAuras.cpp#L1982) kind of strange? I don't have a test environment working on this computer so I can't help.

SinisterMatti:
I'm horde, and I've confirmed with other tauren druids that they're having the same issue.

AndersSahlin:
[https://github.com/cmangos/mangos-classic/blob/master/src/game/SpellAuras.cpp#L1537](mangos) has the same model id for both factions. Not sure if that's right but at least it's better than nothing...

PR #83, opened 2017-03-21
Possible fix for #71

henkethebenke:
Same as cmangos. #71


StadenElysium:
I'm not sure who broke that or why, and without history, I guess I'll never know for sure. The old version looked like this:
```
        case FORM_AQUA:
            if (Player::TeamForRace(target->getRace()) == ALLIANCE)
                modelid = 2428;
            else
                modelid = 2428;
            break;
```

PR #85, opened 2017-03-21
Fix French translation, current one is based on 3.0+

Neo2003:
Current French translation is not Vanilla but from a version 3 or 4 of WoW.
Ex: City names were not translated from English at 1.12 time
Many creatures names are wrong and are from Cataclysm.

Data source: A project I started more than 10 years ago using hundreds of WDB from 1.12 then the project was leaded by a French translation team for many years using bestiary and other sources.

It updates only update French locale and does not touch other translations.

Signed-off-by: Neo2003 <Neo.2003@Hotmail.fr>


m-laniakea:
Good work, this has my approval. There is quite a bit of junk, but it seems those lines won't affect anything.
I will do a bit more testing on in tomorrow and create backups.

spectras:
Sounds great. I might consider installing the French client again. I've been using exclusively English clients since Blizzard team renamed famous places and items, breaking their connection to Warcraft 3 lore.

PR #89, opened 2017-03-22
Fix for transports on MacOS

martinjlowm:
The issue with transports on MacOS is that the client movement info is not updated on `MSG_MOVE_WORLDPORT_ACK`, instead, the information must be sent then to avoid being teleported elsewhere. Sending a heartbeat will inform the client of the new position which is set after a teleport event is scheduled.

Another issue this leads to is the client player camera facing south on a successful port. The movement of the transport object does not affect the camera at all. The client resets the camera in two situations as I've discovered them: Upon leaving a transport object -and- triggering a `SMSG_STANDSTATE_UPDATE` event. This fix sends the event with the current state such that only the camera is reset.

I have tested it with the Windows client and the behavior remains the same as before this change.


martinjlowm:
This patch should be reverted. Apparently, people claim to be stuck in the air.

https://www.reddit.com/r/ElysiumProject/comments/626zwy/zeppelins_are_completely_screwed_for_mac_users/

CrshOverride:
As a Mac user, I can confirm that we either get stuck in the air (with no character loaded) or dumped randomly into the ocean. Unfortunately I don't see anything that screams to me in the changeset that would cause the behavior. I'd be glad to help investigate or test any potential changes.

martinjlowm:
I have recompiled the core with all the changes in `development`-97589d8 (including this patch) and with the following build options and dependencies on MacOS 10.12.3:
```
Build params : "-O3 -DNDEBUG  -std=c++14 --no-warnings -fexceptions -fnon-call-exceptions -march=native -mtune=native -pipe "
Alloc library: TBB/Release
Core Revision: 97589d86bf8cef43ee45 / 2017-03-29 10:42:37 -0700 / MacOSX_x64 (little-endian)
OpenSSL 1.0.2k  26 Jan 2017 (Library: OpenSSL 0.9.8zh 14 Jan 2016)
Using ACE: 6.4.1
MySQL client library: 5.7.17
MySQL server ver: 5.7.17 
```

I have tested this locally by routing my traffic through an AWS EC2 instance in Ireland to increase my latency (I'm based in Denmark) to ~100ms (on Elysium I have 40ms) and I haven't been able to reproduce the behavior as it is on the live realms. The server is hosted on a MacBook Pro and the client runs on a MacBook Air.

@CrshOverride can I reach you on Discord or something? It would be nice of you if you could test the behavior on my server.

CrshOverride:
@martinjlowm My Discord handle is the same as my GitHub handle. Feel free to reach out whenever you have a chance.

PR #95, opened 2017-03-22
instance_ruins_of_ahnqiraj.cpp

shmgiy429:
Assignment within conditional expression, probably meant to check the value of m_auiEncounter not set it.


Rushor:
wrong pr mode /close?

shmgiy429:
Yes, apparently I have no idea how to use GitHub Desktop; sorry new to this workflow... Only meant to change a logical assignment error in one .cpp file and commit that file change to the current list of development changes.

PR #102, opened 2017-03-23
Shift looting quest items investigation and fixes

FractalCabush:
After much investigation, one issue could be this, now changed if statement. Before it checked if qitem existed and only then rebound the pointer for qitem2. However it was ment to check if qitem2 was != nullptr. I have tested locally and seem to be able to shift-loot quest items properly. I am also aware of the fact that this file could be different on the live servers so it needs to be tested.

When shift looting a mob with an item the following happens

- The opcode CMSG_AUTOSTORE_LOOT_ITEM gets sent from the client.
- A handler is dispatched and the mobs items is cross referenced against active quests and if the item to be looted matches one of them. It then calculates how many items are needed, and returns the item to be looted by the client.

However, if player->StoreNewItem() fails then nothing will happen, which we have had multiple reports of.
This can only happen during these conditions

- player->CanStoreNewItem() returns with success
- player->StoreNewItem() fails for some reason.

Another reason for why it might fail is item->AllowedForPlayer() can returns false  and can also cause the item to not loot during shift looting
can also fail if there is no valid qitem pointer AND the item is blocked at the same time. This happens from what i can tell, only when the loot is being generated in the corpse/unit.

Testing and feedback on this is much appreciated.


FractalCabush:
I should prob say that what i have tested is looting quest items from mobs, which i have seen being bugged on elysium. I can confirm that shift-looting gameobjects that contain quest items does not work and will need more investigation

brotalnia:
Did you intend to merge all changes from the Giperion branch into Development, or did you clone the wrong branch?

shmgiy429:
The file changes for LootMgr.cpp look like pretty straightforward corrections to me, but brotalnia is on to something; I see a bunch of unrelated changes as well.

FractalCabush:
I based my local development branch on gyperions stuff because i wanted to have a working build with mmaps

FractalCabush:
i will correct this by closing this pr and opening a new one

Issue #87, opened 2017-03-21
Lokhtos Darkbargainer in BRD does not sell *Plans: Dark Iron Boots* as he should


chevas:
Thorium Brotherhood Rep - A player who is exalted should be able to buy Dark Iron Boots. They are in the game well before the current patch, but the NPC does not have Dark Iron Boots Plans for sale. There are existing players (Lebowski on Anathema Elysium is one of them) that have already learned the plans and currently make the boots. He mentioned that Nano from the original Nostalrius had to fix the same issue for him once he reached exalted with TB.

henkethebenke:
On which server? I guess Anathema since Lebowski plays on that server.

FractalCabush:
The Plans: Dark Iron Boots are available in patch 1.7 so if you are on anathema then yes it should be added to his vendor items.

chevas:
We spoke on Discord, but for Github record:
Anathema on Elysium

Thanks for your response!

m-laniakea:
Confirmed by QA, and admin has been notified.

chevas:
Any update on this being pushed to production?

chevas:
Bump

chevas:
How hard is it to add a database entry? I've been waiting for months to have this item added to the vendor.

m-laniakea:
@chevas Our admins are very busy. Things like this just get lost sometimes.

chevas:
Understood. Apologies for my snide comment.

brotalnia:
I submitted a fix for this two months ago, but they denied it, saying it should be added in itemization, even though the plans are sold by the vendor in 1.12, and to my understanding the base database is supposed to be for 1.12 since there are no itemization patches for 1.12. They even have queries to remove it from the vendors in some of the itemization patches, even though it is not there in the default db, which means the queries do nothing. https://github.com/elysium-project/itemization/search?utf8=%E2%9C%93&q=20040&type=

At the same time, they make changes that are not supposed to be in 1.12 as migrations, instead of in the itemization patches, where patch specific changes are supposed to go. https://github.com/elysium-project/server/commit/548dee68edb5bf2afc3b2a66fe4c469c210a2fc4

So i am completely confused as to what is supposed to be in the database and what is not anymore. There needs to be a post explaining what changes are to be made in migrations, and what in itemization, and most importantly what patch is the default database for -> https://github.com/elysium-project/server/releases/tag/02.03.2017

chevas:
[...]

PR #108, opened 2017-03-24
Add 'Plans: Dark Iron Boots' to vendor list for 'Lokhtos Darkbargainer'

brotalnia:
Fixes issue https://github.com/elysium-project/server/issues/87

The plans are not present in the npc_vendor table at all, even though in the itemization patches there is a query to remove them in patches 1.2 and 1.3 - https://github.com/elysium-project/itemization/search?utf8=%E2%9C%93&q=20040&type=

Since the "base" world database is supposed to be for 1.12.1, they should be in the vendor list.

### DIFF
--- a/sql/migrations/20170324161850_world.sql
+++ b/sql/migrations/20170324161850_world.sql
@@ -1,0 +1,4 @@
+INSERT INTO `migrations` VALUES ('20170324161850');
+
+-- Add "Plans: Dark Iron Boots" to "Lokhtos Darkbargainer".
+INSERT INTO `npc_vendor` VALUES ('12944','20040','0','0');


m-laniakea:
If your SQL query is patch-specific, please create a request here: https://github.com/elysium-project/itemization/tree/master/sql/1.07

brotalnia:
I asked Shenna on discord and she says if something is missing from the 1.12 database ( https://github.com/elysium-project/database ) it should be added as a migration and not as an itemization patch. It doesn't make sense to add it to the 1.7 itemization patch, because then when the server moves to 1.8 and they put back 1.12 database and patch it with the 1.8 itemization patch it, the item will be gone again!

What needs to be added to itemization patches 1.5 and 1.6 is a query to remove the item from the vendor list because it is not out yet, just like in the 1.2 and 1.3 patches. But currently that is pointless because it's missing from npc_vendor in the DEFAULT 1.12 database ( https://github.com/elysium-project/database ), so it's trying to remove something that was never there in the first place. My pull request is to add it to the default database as it is missing from there, and that is a bug.

As you see it is there in the MangosZero database - https://github.com/mangoszero/database/blob/master/World/Setup/FullDB/npc_vendor.sql#L11349

PR #131, opened 2017-03-27
The Woodland Protector - quest completion text (FIX)

FractalCabush:
corrected the quest text for the woodland protector.

PR #132, opened 2017-03-27
Fandral Staghelm reffered to as "ArchHunter" (FIX)

FractalCabush:
Fixed several places in the npc quest text where it was referring to $C instead of Druid/druid.


m-laniakea:
Just update the other ones I mentioned here in one go, please.
https://github.com/elysium-project/server/pull/131

FractalCabush:
What do you mean in one go? i took everything referred in the github issue

brotalnia:
He means make the changes for the others in 1 pull request instead of opening a new one for each one because the problem is the same.

PR #133, opened 2017-03-27
Archmage Xylem not interacting (FIX)

FractalCabush:
Changed the npcflag to include gossip for Archmage Xylem, also conneted the npc_text gossip to the actual npc_gossip table.


FractalCabush:
As far as i could see those quests were not available, since they are added in patch 1.6. I wasnt able to check on 1.6 if he gave the correct one for your specific class or all.

bdebaere:
This is correct?
SELECT * FROM npc_text WHERE id = 8379 gives me a text for Burning Crusade:
"Isn't this just so much fun? I'm so glad that Lord Saltheril throws these parties every night, especially with all of that nastiness occurring to the south. Kind of just makes all of that go away, don't you think? "

PR #176, opened 2017-03-28
Corrects the errors in quest texts relating to Fandral Staghelm

FractalCabush:
Also known as the Arch Druid. This fixes the quest texts for quest: 933, 934, 940, 952 and 3785.


m-laniakea:
@FractalCabush Please don't use `world`. Others might have a different name for their world database.

FractalCabush:
You are so right! My bad! Sorry

Skickat från min iPhone

Issue #183, opened 2017-03-29
Bloodfang proc (8/8 rogue Tier 2) self buff part can be dodged or parried by the rogue.


Slicy:
Bloodfang proc (8/8 rogue Tier 2) is composed of a damage part on the target and a "buff" part for the rogue.

When the proc occurs, the rogue will sometimes DODGE or PARRY the buff part. It seems the rogue's own dodge/parry chance are taken into account while it shouldn't.

Example : http://imgur.com/a/Ovwrk

The Bloodfang buff should apply to the rogue with no chance of failure. Each time Bloodfang procs, only the damage component can be dodged/blocked/parried by the target. The HoT self buff part should never be dodged/parried by the rogue nor the target.

The Bloodfang proc has 2 effects :

Spell Details

Duration 6 seconds School Physical Mechanic n/a Dispel type n/a

Effect #1 (2) School Damage (Physical) Value: 283

-> This is the damage part of the proc.

Effect #2 (6) Apply Aura #8: Periodic Heal Value: 50 Interval: 1 seconds

-> This the self-healing buff part of the proc.

Unknown-as-Captain:
Haha, funny

PR #188, opened 2017-03-29
Prevent Bloodfang proc buff being wrongly Parried or Dodged

brotalnia:
Fixes issue - https://github.com/elysium-project/server/issues/183

This makes it so the buff cannot be parried or dodged by the rogue.

bloodfang proc fix
Tested with 100% dodge chance.


Slicy:
Thank you.

Issue #194, opened 2017-03-29
Raid markers disappear when one of the marked mobs is killed: CONFIRMED


QAkvastskaft:
BUGGED BEHAVIOR
I have noticed that the raid markers will often disappear when one of the marked mobs is killed. For example, if mobs are marked with a skull, X, and moon and any of those mobs are killed, the other symbols above the remaining mobs disappears immediately. This occurs both outside in the open world, and in dungeons. I have ruled out an addon problem by disabling all of them and I have confirmed this issue with others experiencing it.

 CORRECT BEHAVIOR
The correct behavior should be that when one of the marked mobs is killed, the remaining mobs retain their marked symbols. They should not disappear when one of the mobs is killed. Please advise on if this is a known bug and if it will be fixed.

 SOURCES
My own experience and the experiences of other players who have seen this bug.

Original post on bug tracker here: https://elysium-project.org/bugtracker/issue/567

QAkvastskaft:
Fixed: https://forum.elysium-project.org/topic/47461-development-update-24052017/

PR #205, opened 2017-03-29
Fix to raid markers disappearing (#194)

GuybrushGit:
https://github.com/elysium-project/server/issues/194

This issue happens because the UpdateLooterGuid method triggers a SendUpdate().
It seems like every time a _SMSG_GROUP_LIST_ message is sent to a non-raid group, the client is expecting the raid targets to be sent as well. Since none are ever sent the raid markers get cleared.

Ideally we would want to send the raid markers in the _SMSG_GROUP_LIST_ message, but I tested appending them to the end of the packet and it wasn't working. We would need to find out the exact structure of the _SMSG_GROUP_LIST_ packet.

The next best alternative is to send the target icons list to the player right after the SendUpdate().
It seems unecessary to do this on every SendUpdate call, so I did it only for the UpdateLooterGuid method.

### DIFF
--- a/src/game/Group/Group.cpp
+++ b/src/game/Group/Group.cpp
@@ -1966,4 +1966,17 @@ void Group::UpdateLooterGuid(WorldObject* pLootedObject, bool ifneed)
         SetLooterGuid(0);
         SendUpdate();
     }
+
+    // SendUpdate clears the target icons, send an icon update
+    if (!isRaidGroup()) 
+    {
+        for (member_citerator citr = m_memberSlots.begin(); citr != m_memberSlots.end(); ++citr)
+        {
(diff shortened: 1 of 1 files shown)


Unknown-as-Captain:
HE ACTUALLY DID IT

THE ABSOLUTE MADMAN

StadenElysium:
I'll investigate the packet structure when I get a moment and report back.

Tenarie:
This takes the #1 price.

elysium-duks:
Reviewed the packet structure - falls in line (atleast from what I see)

Seagulls:
Can this get committed already?

Unknown-as-Captain:
Thank you LOTS once again for this fix
Gotta wait for the review process I suppose

PR #208, opened 2017-03-30
Fixed Blackout proc conditions

GuybrushGit:
Issues:
https://elysium-project.org/bugtracker/issue/1831
https://elysium-project.org/bugtracker/issue/1682

This is currently being abused in pvp to interrupt flag captures from across the map.
Blackout should only proc on shadow spells that deal damage.


sql/migrations/20170330135457_world.sql
```diff
@@ -0,0 +1,9 @@
+INSERT INTO `migrations` VALUES ('20170330135457');
+
+-- Set blackout to only proc on shadow damage spells
+DELETE FROM `spell_proc_event` WHERE `entry` IN (15268,15323,15324,15325,15326);
+INSERT INTO `spell_proc_event` (`entry`,`SchoolMask`,`procFlags`) VALUES (15268,32,65536);
+INSERT INTO `spell_proc_event` (`entry`,`SchoolMask`,`procFlags`) VALUES (15323,32,65536);
```
brotalnia: This produces the following errors when loading the database:

```
ERROR:Spell 15323 listed in `spell_proc_event` is not first rank (15268) in chain
ERROR:Spell 15324 listed in `spell_proc_event` is not first rank (15268) in chain
ERROR:Spell 15325 listed in `spell_proc_event` is not first rank (15268) in chain
ERROR:Spell 15326 listed in `spell_proc_event` is not first rank (15268) in chain
```

A comment in the code says it is not being loaded, as it will have no effect.
https://github.com/elysium-project/server/blob/8dca0a87c1c4a9204c13042ac13c0071a5230cde/src/game/Spells/SpellMgr.cpp#L1270

PR #217, opened 2017-03-31
SpellAuras: fixing shapeshifting + transform and stacked transforms (#245)

WobLight:
- transform:
    - ApplyModifier checks if the aura is applied, which is always true for transforms effects, use directly HandleAuraTransform to skip the checks.
    - When removing last transform, reapply shapeshifting (if any).

- shapeshift:
    - use Real guard to split ability and display effects

I don't have any knowledge of mangos, but using handles directly looks like a workaround to make another workaround working to me. Well at least now it should work.

### DIFF
--- a/src/game/Spells/SpellAuras.cpp
+++ b/src/game/Spells/SpellAuras.cpp
@@ -2218,7 +2218,7 @@ void Aura::HandleAuraTransform(bool apply, bool Real)
     if (apply)
     {
         // Discombobulate removes mount auras.
-        if (GetId() == 4060)
+        if (GetId() == 4060 && Real)
             target->RemoveSpellsCausingAura(SPELL_AURA_MOUNTED);
         if (GetId() == 23603)   // Ustaag <Nostalrius> : Nefarian Class Call Mage
         {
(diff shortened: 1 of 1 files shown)


WobLight:
I'd like to point out that either the stacked transform case was implemented but untested or the `IsApplied() == apply` check in `ApplyModifier` was added afterwards. Also, I'm not quite sure what the exact purpose of Real flag is. If the check is legit, then a way to avoid direct usage of `HandleAuraTransform` and `HandleAuraModShapeshifting` would be: unapply (not Real) the shapeshift/transform when being overridden, then reapply (not Real) and unapply (Real) when removed or reapply (not Real) when no longer overridden.

WobLight:
There's a problem with tauren scaling, I'm working on fixing it.
I'm also assuming tauren scaling should be normalized when transformed.

WobLight:
I've checked on mangoszero, indeed the `IsApplied() == apply` guard in `ApplyModifier` is not there. In my opinion, it conflicts with the `Real` logic and is worthy of further discussion.

WobLight:
Everything should be working now.

WobLight:
I'm also concerned by  
```c++
// and polymorphic affects
if (target->IsPolymorphed())
    target->RemoveAurasDueToSpell(target->getTransForm());
```
as far as I can tell, this code will remove only mage polymorph, and only if it is the last negative transform applied. Furthermore, even commenting these out, shapeshifting will still remove all polymorph effects correctly, so it's probably duplicated code.

WobLight:
referencing follow up #440

PR #232, opened 2017-04-02
Fixes transportation issues on MacOS #2

martinjlowm:
This pull request is a follow-up to #89 which appeared to bring back the 'old' horrors of random teleportation in production.

#89 didn't cover the situation with dedicated threads for packet broadcasting (`Network.PacketBroadcast.Threads`). In addition to that, the heartbeat "self" flag should have been set to true. The heartbeat was being passed to the client without any dedicated threads because of `Object.cpp:1882` (`SendObjectMessageToSet(&data, true, except);`) which forced the "self" flag to be true. This led to a scenario that worked despite having the "self" flag set to false until... `Network.PacketBroadcast.Threads` was configured differently in production.

With broadcast threading, there is a small chance that the heartbeat packet can be skipped (if the outgoing packet queue is full) and cause the random teleportation. The change of conditions in `CanSkipPacket` attempts to avoid this.

Camera is still an issue, and is now handled upon `HandleZoneUpdateOpcode` instead. `HandleMeetingStoneInfoOpcode` fired too early and only worked 60% of the time.

### DIFF
--- a/src/game/Handlers/CharacterHandler.cpp
+++ b/src/game/Handlers/CharacterHandler.cpp
@@ -795,15 +795,6 @@ void WorldSession::HandleMeetingStoneInfoOpcode(WorldPacket & /*recv_data*/)
     WorldPacket data(SMSG_MEETINGSTONE_SETQUEUE, 5);
     data << uint32(0) << uint8(6);
     SendPacket(&data);
-
-    // Trigger a client camera reset by sending an `SMSG_STANDSTATE_UPDATE'
-    // event. See `WorldSession::HandleMoveWorldportAckOpcode'.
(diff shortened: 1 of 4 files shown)


StadenElysium:
Just to note, the packet broadcaster is disabled on the realms.

martinjlowm:
Okay! That is strange. It was just an assumption since it had an immediate effect when I turned it on. In either case, the change of the "self" flag should _hopefully_ do the trick.

If this also fails in production, can you confirm that heartbeats are actually sent to the affected players? I see there's a handy unit debugging command available for GMs to trace packets.

martinjlowm:
2000 online, the patch works so far

PR #233, opened 2017-04-02
Extend pinfo to display 2FA status, gold in mail inbox and gold in ma…

StadenElysium:
…il outbox


darfink:
I believe you forgot `INSERT INTO `migrations` VALUES ('20170402093309');`

PR #248, opened 2017-04-02
Use Spell Focus as effect for Spirit of Aquamentas

darfink:
The item [*Spirit of Aquamentas*](http://db.vanillagaming.org/?item=11904) does not function as described by the tooltip.

image

Instead it increases the damage and healing of spells by 20.
The misinformation provided by the tooltip causes much confusion (see among others [#2792](https://elysium-project.org/bugtracker/issue/2792)).

### DIFF
--- a/sql/migrations/20170402195949_world.sql
+++ b/sql/migrations/20170402195949_world.sql
@@ -1,0 +1,4 @@
+INSERT INTO `migrations` VALUES ('20170402195949');
+
+-- Change Spirit of Aquamentas to decrease spells mana costs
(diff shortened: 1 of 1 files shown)


StadenElysium:
Patch 1.3: The effect of the Spirit of Aquamentas and Choker of Enlightenment has changed to a +damage and +healing of all spell schools effect.

brotalnia:
Shouldn't this change be added to the 1.2 itemization patch then?
https://github.com/elysium-project/itemization/blob/master/sql/1.02/1.02_pve_items.sql

PR #258, opened 2017-04-03
Code cleanups

WobLight:
Just a minor legibility improvement.


Kittnz:
why would this be any improvement.. 4 spaces is more then enough

WobLight:
spaces? I removed useless `if`s, probably compiler is smart enough to optimize it on its own, but those long `if`s are really harder to read then a proper return with boolean operators.

Kittnz:
Oh my bad, didn't look well enough apparently. Carry on 👍

Issue #267, opened 2017-04-04
Auction House search not returning all results


brendan98:
I'm on horde side on Elysium PvP server.  When you search for items of a type and within a level range you are not seeing everything available. 

for example, I was searching for cloth chest pieces level 50-60 hoping to find a nice blue robe and never saw anything I wanted.  Then I searched explicitly (by name) for "robe of winter night", and there it was!!  It didn't come up when I searched by category but it did when I searched by name.

I've seen this happen on other items too, but can't remember which ones specifically.  Easily repeatable bug.

StadenElysium:
Was able to reproduce this. The robe shows up when you select 'cloth' but not when you select the 'chest' category.

PR #279, opened 2017-04-05
fix for issue #267

WobLight:
Chest slot has two inventory types, explicit check for `INVTYPE_ROBE` is thus necessary.


WobLight:
referencing #267

Issue #192, opened 2017-03-29
Quest items in the bank: CONFIRMED


QAkvastskaft:
BUGGED BEHAVIOR
I placed a Fungal Spore, for the quest Fungal Spores, a Raptor head for Raptor Thieves, and two Prowler Claws for the Quest Prowlers of the Barrens, all in the Barrens in my bank to do a dungeon, afterwords I grabbed them and wen't back to finish the quests, but the items are not dropping, killed 10 prowlers without getting one stuck at 5/7, 12 raptors stuck at 11/12 and cannot click on the last Fungal Spore in either Stagnant Oasis, Lushwater Oasis or The Forgotten Pools. Also the quest givers have question marks above their heads as if I can complete the quest

 CORRECT BEHAVIOR
The Spore should be lootable for sure, the others I could just be extremely unlucky, my xloot says the prowlers only have a 20% chance to drop right now from all that I've killed.

 SOURCES
Links to the quest givers question marks and un-clickable fungal spores Can't see my cursor in the screenshot, but its over the Spore

http://imgur.com/a/KwiKB

COMMENTS from QAKvastskaft:

To reproduce: .go creature william pestle
take quest kobold candle
.itemadd 772 4
(Or .go -9848 238.6 21.1 0 and kill kobold until 4/8 )
.go creature olivia burnside
bank the quest items (and take them out again, I guess, can't remember :))
.go -9848 238.6 21.1 0
kill kobold, no items will now drop despite still only being 4/8 candles.
Abandoning the quest and retaking it will work, this will not remove the quest items. In this case .quest add 60

Re-logging does not fix issue.
Restarting the client does not fix the issue.

Original post on bug tracker here: https://elysium-project.org/bugtracker/issue/259

Unknown-as-Captain:
Secondhand confirmed, very annoying and very important bug.

Great steps to reproduce!

WobLight:
so do I delete my pull request I guess?

QAkvastskaft:
Fixed: https://forum.elysium-project.org/topic/44683-development-update-19042017/

PR #287, opened 2017-04-05
quickfix for issue #192

WobLight:
With this, everything should work fine. (#192)
However, there are following considerations:
1. I've assumed that quest items should counted in quest log even while banked
2. Instead of completely removing ItemAddedQuestCheck, I've changed the item count to 0, so the client will still be notified (it will display an alert even though the item count is not changed)

Both 1. and 2. can be easily fixed if they're not intended to be.


WobLight:
Merged in pull request #45, closing

WobLight:
Nevermind, got confused by github referencing.

PR #290, opened 2017-04-06
Improve regen accuracy

StadenElysium:
Changed the timing for health/mana/energy/rage generation to allow negative values. Any negative value will effectively be subtracted from the time until the next regeneration tick, thus reducing drift/error over time, especially on servers with higher load.

This should also fix an issue with Evocation reported at https://elysium-project.org/bugtracker/issue/824


StadenElysium:
Should also make some rogue energy tick addons more accurate (e.g. EnergyWatch).

Issue #260, opened 2017-04-03
[Battleground] Remove Insignia: players killed by NPC cannot be removed


WobLight:
Player killed by NPCs in Alterac Valley cannot have their insignia removed until they do "release spirit".

QAkvastskaft:
Fixed here: https://forum.elysium-project.org/topic/44683-development-update-19042017/

PR #302, opened 2017-04-06
[Battleground] fix player death handling

WobLight:
Also fixes issue #260


QAkvastskaft:
Fixed here: https://forum.elysium-project.org/topic/44683-development-update-19042017/

Issue #65, opened 2017-03-19
Herod spawns multiple adds upon being pulled, lever does not close door, adds do not despawn on Herod's death


Unknown-as-Captain:
Remake of https://elysium-project.org/bugtracker/issue/2700 as the Elysium Project bugtracker will soon be deprecated.

## Expected Behavior
Herod has an anti-kiting mechanic that spawns 3 (_three_) Scarlet Myrmidons if you try to kite him out of the room. These adds **only** spawn if Herod moves outside the room, otherwise, they should not spawn.

## Bugged Behavior
Herod currently always spawns 4 (_four_) Scarlet Myrmidons immediately upon being pulled. They always aggro as soon as they spawn, no way to avoid them.
Examining the script for this bossfight shows that [you should be able to prevent the adds from aggroing by having the door closed when pulling Herod](https://github.com/elysium-project/server/blob/19e7622bd3f3d2c0b1bfe65892f25d035b9e222b/src/scripts/eastern_kingdoms/tirisfal_glades/scarlet_monastery/boss_herod.cpp#L111), however, currently the lever to close the door is not working either, so you can't close the door to prevent these adds from aggroing at all.
Furthermore, [the script also shows that these adds are supposed to despawn on Herod's death](https://github.com/elysium-project/server/blob/19e7622bd3f3d2c0b1bfe65892f25d035b9e222b/src/scripts/eastern_kingdoms/tirisfal_glades/scarlet_monastery/boss_herod.cpp#L157), this is currently not working on the live server, the adds remain alive. (Unfortunately I don't know if this is blizzlike behavior or not, I was unable to find any videos of the Myrmidons spawning during Herod's fight, the anti-kiting mechanic was extremely obscure during Retail as it was rarely relevant)

So this is a weird issue where the server's script has some weird not-blizzlike mechanic centered around the door, which, unfortunately, is also bugged. Even if it wasn't bugged, there's still a few issues with it: You could simply close the door, pull Herod, then open the door and kite him out of the room with no adds to stop you. Also, since this is a custom mechanic (not blizzlike), many players simply do not expect it to happen, closing the door was never a part of the scripting in Retail, so players never learned to do it.

## Sources
Wowpedia details the anti-kiting mechanic: http://wow.gamepedia.com/Herod#Strategies
>Also, if you try to run outside the room, Herod will sprint, increasing his movement speed by 60% and almost certainly catching up to you; if you don't get affected by this, 3 Scarlet Myrmidons will spawn and block your path.

[...]

PR #304, opened 2017-04-07
Proposal fix for Herod's savage Myrmidons spawn on aggro

sipertruk:
~~The spawn happens now only when Herod's target exits his room.
They despawn on Herod's reset, they don't despawn on his death and still drop nothing.~~
The spawn happens now at the opposite side of the corridor before the room.
They will engage a player coming close or Herod's target when he exits his room, running toward it.
They always despawn on Herod's reset, they despawn on his death only if they have no target, they drop the usual stuff.

I couldn't find any problem with the door and lever.

Related reports and duplicate
Issue #65
https://elysium-project.org/bugtracker/issue/2700
https://elysium-project.org/bugtracker/issue/3900
https://elysium-project.org/bugtracker/issue/3732
https://elysium-project.org/bugtracker/issue/2950


FractalCabush:
Nice! However its not when the target leaves the room but when herod leaves the room ( following his target ofc) that they should spawn :)

m-laniakea:
Good. Line 171, just change that to `!m_creature->IsWithinDist2d(RoomCenter.x, RoomCenter.y, 32.0f)`. Perhaps give a bit more leeway as well, as distance calculations are not guaranteed to be accurate.

sipertruk:
Is it to ensure a safer result of IsWithinDist2d or do you have precise info ?
If the player exits the room and run down the corridor they would spawn behind him coming from nowhere, it's a less rational event.
In fact we better spawn them at the other end of the corridor, make them wait and aggro only when herod exits or despawn on his death/reset.
That would fit what is said here, just found it, but it's from 2007 : http://wowwiki.wikia.com/wiki/Talk:Herod

sipertruk:
Updated my 1st comment, maybe 32.0f is fine, it's already a bit further than any place in the room.

Unknown-as-Captain:
Seems to fix everything that was wrong in the fight, I love this.

We might never get 100% certain details of what the "blizzlike" behavior was, but given the incredible obscurity of this anti-kiting feature (that 99% of people won't even see, ever) this is a working, more-than-acceptable implementation.

What's important is that the door isn't important anymore, and that the adds shouldn't even be noticed by the players unless they run out to the hallway. This fix does that perfectly.

PR #309, opened 2017-04-07
Arcanite Dragongling fix

Jonathanfy:
based  on
https://elysium-project.org/bugtracker/issue/2644

Ish fixed. 

Added a script for flame attack and modified values for it's attack.

Chance on hit is mostly guesswork same with the spells used. They seem to be around the values mentioned.

### DIFF
--- a/src/scripts/world/npcs_special.cpp
+++ b/src/scripts/world/npcs_special.cpp
@@ -1407,15 +1407,74 @@ CreatureAI* GetAI_npc_gnomish_battle_chicken(Creature* pCreature)
     return new npc_gnomish_battle_chickenAI(pCreature);
 }
 
+
+/*######
+## arcanite dragonling
+######*/
+enum
+{
+	SPELL_Flame_Buffet = 25651,
+	SPELL_Flame_Breath = 20712
+};
+
+struct npc_arcanite_dragonling_dragonlingAI : ScriptedPetAI
+{
+	explicit npc_arcanite_dragonling_dragonlingAI(Creature* pCreature) : ScriptedPetAI(pCreature)
+	{
+		m_creature->SetCanModifyStats(true);
+
+		if (m_creature->GetCharmInfo())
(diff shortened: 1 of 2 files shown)


src/scripts/world/npcs_special.cpp
```diff
@@ -1407,15 +1407,74 @@ CreatureAI* GetAI_npc_gnomish_battle_chicken(Creature* pCreature)
...
 
+
+/*######
+## arcanite dragonling
+######*/
 enum
 {
-    SPELL_IMMUNITY      = 29230,
-    SAY_CLEANER_AGGRO   = -1289010
+	SPELL_Flame_Buffet = 25651,
```
Kittnz: tabs

StadenElysium:
Apart from needing to switch tabs to spaces, it's looking good. Thanks!

GuybrushGit:
None of the sources show the pet doing a Flame Breath for 900 dmg, that should be removed.
Also the Flame Buffet you used is an instant cast, it should be this one instead:
http://db.vanillagaming.org/?spell=9658

Jonathanfy:
hi, the 900 damage shouldn't have been picked. :). From the sources I gathered the dragon is supposed to do a breath and at the same time apply a fire buffet. So the breath should do about ~200 and then the firebuffet.

Jonathanfy:
Ok, removed the flame breath. Only problem is that the fire buffet is not doing the correct damage. According to sources it should do between 100 and 300 damage. (differs between sources). '

brotalnia:
If this spell is not used anywhere else, you could change the damage in the `spell_effect_mod` table in the database. Or you could use CastCustomSpell instead to specify custom damage.

Jonathanfy:
castcustomspell sounds like the better alternative.

Jonathanfy:
Now it should work as intended. A 2 sec cast flame buffet and a random flame breath that does 300 damage.

PR #318, opened 2017-04-08
Visibility modifiers - WIP

StadenElysium:
WIP for feedback - do not merge.


Hiperzone:
Are you sure assigning a custom field to the mirror object descriptors will work?
Usualy these fields are quite sensitive to changes and WoW expects them to have a certain size otherwise you get the typical 132 error.

PR #324, opened 2017-04-08
threads improvement for MapManager::Update

WobLight:
Warning: concurrency involved, please review carefully.

### DIFF
--- a/src/game/Maps/MapManager.cpp
+++ b/src/game/Maps/MapManager.cpp
@@ -241,33 +241,21 @@ void MapManager::DeleteInstance(uint32 mapid, uint32 instanceId)
 class MapAsyncUpdater : public ACE_Based::Runnable
 {
 public:
-    MapAsyncUpdater(bool* updFinished, uint32 updateDiff) :
-        updateFinished(updFinished), diff(updateDiff), loops(0)
+    MapAsyncUpdater(uint32 updateDiff) :
+        diff(updateDiff)
     {
     }
 
     virtual void run()
     {
         WorldDatabase.ThreadStart();
-        do
-        {
-            for (std::vector<Map*>::iterator it = maps.begin(); it != maps.end(); ++it)
-            {
-                if (loops && *updateFinished)
-                    break;
-                (*it)->DoUpdate(diff);
-            }
-            if (!(*updateFinished))
-                ACE_Based::Thread::Sleep(5);
-            ++loops;
-        }
-        while (!(*updateFinished));
+        for (std::vector<Map*>::iterator it = maps.begin(); it != maps.end(); ++it)
+            (*it)->DoUpdate(diff);
         WorldDatabase.ThreadEnd();
     }
     std::vector<Map*> maps;
-    volatile bool* updateFinished;
-    uint32 diff;
-    uint32 loops;
+private:
+    const uint32 diff;
 };
 
 class ContinentAsyncUpdater : public ACE_Based::Runnable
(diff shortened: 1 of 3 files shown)


WobLight:
more work coming soon

WobLight:
Added a new `ThreadPool` class to avoid spawning new threads at every update.
It's general-purpose and makes use of c++11/14 and std libraries (std::thread, std::async, ...).

WobLight:
if we have use cases for this, we can make `operator<<` threadsafe and change std::stack to std::queue to avoid starvation.

WobLight:
some basics tests show performances are not good (probably due to locking to pick from the queue), I'll work on improving them.

Hiperzone:
to be honest, getting ride of ACE would be the best approach and use boost asio instead.

WobLight:
I'm using std indeed

WobLight:
`ThreadPool` overhead should now be lower than current ACE implementation

martinjlowm:
Nice work! I am really looking forward to see the end result. In fact, I have changes ready to fully support MacOS and am just waiting for this to be stable and merged into production. Right now, ACE thread (pthread) identifiers are stored as struct pointers, rather than a 32-bit integer identifier. This means, wherever the thread id is used, Mac OS needs a specific function call to get an unsigned 64-bit integer of that thread. The end result would be an ugly hack...

WobLight:
I don't get what you are saying, std::thread is a pthread on linux.

martinjlowm:
Yea, `std::thread::id` should be a portable type instead of the inconsistency I mentioned :)

Itzamnaaj:
Thanks for your work on the tools, WobLight.

Sadly, your branch does not compile on GNU/Linux:

`/home/itzamna/src/server/contrib/vmap_extractor/vmapextract/dbcfile.cpp:20:31: fatal error: libmpq\mpq_libmpq.h: No such file or directory`

This is due to backslashes in include paths in several source files.

I fixed this and other compiler errors in my branch, see https://github.com/elysium-project/server/pull/990.

WobLight:
Guess I forgot to test `USE_EXTRACTORS` after last merge.
Now everything should be fine

Itzamnaaj:
Thank you, the tools did compile. However, I have the same problem as in my branch: the tools do compile, but the binaries do not work correctly. It cannot find most models.

`Extracting GameObject models...
Extracting World\Wmo\Kalimdor\Buildings\Nightelfmoonwell\Nightelfmoonwellornate.wmo
No such file World\Wmo\Kalimdor\Buildings\Nightelfmoonwell\Nightelfmoonwellornate.wmo.
Couldn't open RootWmo!!!
Extracting World\Wmo\Playerhousing\Human\Humanlevelonetest.wmo
No such file World\Wmo\Playerhousing\Human\Humanlevelonetest.wmo.
Couldn't open RootWmo!!!
Extracting World\Wmo\Azeroth\Buildings\Stormwind\Stormwind.wmo
No such file.`

`Could not find file of model World\Azeroth\Zulgurub\Passivedoodads\Ruins\Zulgurubruins04.m2
Could not find file of model World\Azeroth\Zulgurub\Passivedoodads\Ruins\Zulgurubruins05.m2
Could not find file of model World\Azeroth\Zulgurub\Passivedoodads\Ruins\Zulgurubruins06.m2
Could not find file of model World\Azeroth\Zulgurub\Passivedoodads\Trees\Zulgurubtree01.m2
Could not find file of model World\Azeroth\Zulgurub\Passivedoodads\Trees\Zulgurubtree02.m2
Could not find file of model World\Azeroth\Zulgurub\Passivedoodads\Trees\Zulgurubtree03.m2
Could not find file of model World\Azeroth\Zulgurub\Passivedoodads\Trees\Zulgurubtree04.m2
Could not find file of model World\Azeroth\Zulgurub\Passivedoodads\Trees\Zulgurubtree05.m2`

The errors go on for a while...

WobLight:
closing as it's not longer updating correctly. New pr #3247

PR #337, opened 2017-04-08
Fix neutral faction pets aggressivity

sipertruk:
Add the missing case of "player versus neutral reputation based factions" to Unit::GetReactionTo. 

Before it was down to returning default faction reactions to the pet's target selection function.

A little bit of clearing too.


sipertruk:
What went wrong ?

Issue #328, opened 2017-04-08
Divine Shield bug, taking damage from ingame fireplaces around the world!!!!!!: CONFIRMED


QAkvastskaft:
BUGGED BEHAVIOR
Farming herbs in blackrock mountain, I see black lotus near dragon elites and go there, DIVINE SHIELD, and start to loot the lotus but halfway through fire damage from the fire place on the ground interrupts me, while I have divine shield. Turns out with divine shield on you take 20 damage per 3-4 seconds from fireplaces.!!!!

 CORRECT BEHAVIOR
Divine Shield should block any damage and everything except Divine Intervention!

 SOURCES
Happened on Anathema PvP on saturday, 21.01.2017

Original post on bug tracker here:  https://elysium-project.org/bugtracker/issue/1965

QAkvastskaft:
Fixed here: https://forum.elysium-project.org/topic/44683-development-update-19042017/

PR #342, opened 2017-04-08
Allow Environmental Fire Damage to be Absorded or Resisted

brotalnia:
Fixes issue - https://github.com/elysium-project/server/issues/328

This makes it so environmental fire damage can be absorbed or resisted, like that from campfires. 

fire dmg absorbs

I don't know if it is supposed to be absorbed, or just do nothing if you are immune though. Could just change the if in the beginning of the function to prevent all damage when immune:

```
if (!IsAlive() || isGameMaster() || this->IsTotalImmune())
{ return 0; }
```


thepparker:
It should either say you're immune to the damage, or you absorb all the damage, similar to other types of attacks when DS or a PW:S is up. I think taking a tick of 0 damage may still interrupt casts as well

brotalnia:
It does not interrupt you if you absorb the damage, i tested it.

PR #356, opened 2017-04-10
Fix for the Elemental Charms from Arathi Highlands

brotalnia:
The elemental exiles in Arathi Highlands drop charms, which are supposed to kill the opposite kind of elemental. They had their targets set correctly in the database, but the effect was missing so they were bugged. This fixes them.

http://db.vanillagaming.org/?item=4479
http://db.vanillagaming.org/?item=4480
http://db.vanillagaming.org/?item=4481


brotalnia:
I was working on clearing more of the database errors when i noticed some are related to these items.

> ERROR:Table `item_required_target`: Spell used by item 4479 does not have implicit target TARGET_CHAIN_DAMAGE(6), TARGET_DUELVSPLAYER(25), already listed in `spell_script_target` or doesn't have item spelltrigger.
> ERROR:Table `item_required_target`: Spell used by item 4480 does not have implicit target TARGET_CHAIN_DAMAGE(6), TARGET_DUELVSPLAYER(25), already listed in `spell_script_target` or doesn't have item spelltrigger.
> ERROR:Table `item_required_target`: Spell used by item 4481 does not have implicit target TARGET_CHAIN_DAMAGE(6), TARGET_DUELVSPLAYER(25), already listed in `spell_script_target` or doesn't have item spelltrigger.

Seems like if the spell has targets defined in `spell_script_target`, the item that uses the spell does not need to have target data. Added another commit to resolve the errors.

PR #357, opened 2017-04-10
Sweeping strikes fixes.

Tenarie:
As i've written in the comments, it's a really complicated fix (Atleast in my eyes)
Reason being is the UnitAuraProcHandler isn't made to support Whirlwind or Cleave interaction with sweeping strikes in any way.

This gives a minor bug i really can't wrap my head about how to fix with whirlwind and cleave.
If there is only 1 charge left and you proc sweeping strikes with whirlwind or cleave only one target will be hit by the sweeping strike proc (Which would also happen on live)

Fixed so that Whirlwind taking only one charge.
Fixed so that Cleave taking only one charge.
Fixed Whirlwind not consuming a charge if it's too close to another NPC without hitting it.
Fixed Retaliation to work with sweeping strikes.
Fixed sweeping strikes procs getting double bonus from percentage damage increase. Tested with enrage and death wish (AUCH!)
Adjusted whirlwind range to the same as sweeping strikes.

I'd like some feedback on the variable naming, commenting and code that should be added if something should be what nullified ?? remove variables ?? (whatever these terms is called) since c++ is my first object oriented programming language (i've only worked with bash and powershell scripts).

### DIFF
--- a/src/game/Spells/Spell.cpp
+++ b/src/game/Spells/Spell.cpp
@@ -3190,7 +3190,7 @@ void Spell::cast(bool skipCheck)
 		}
 
 
-		if (m_spellInfo->Id == 20569)                   //Cleave
+		if (m_spellInfo->Id == 20569 || m_spellInfo->Id == 845 || m_spellInfo->Id == 7369 || m_spellInfo->Id == 11608 || m_spellInfo->Id == 11609)  //Cleave all ranks
 		{
 			if (m_caster->GetAura(12292, EFFECT_INDEX_0)) //Sweeping Strike
 			{
--- a/src/game/Spells/Spell.cpp
+++ b/src/game/Spells/Spell.cpp
@@ -3159,55 +3159,54 @@ void Spell::cast(bool skipCheck)
 
             break;
         }
-     	case SPELLFAMILY_WARRIOR:
-	{
-		if (m_spellInfo->Id == 1680)                   //Whirlwind
-		{
-			if (m_caster->GetAura(12292, EFFECT_INDEX_0)) //Sweeping Strike
-				{
+	case SPELLFAMILY_WARRIOR:
+        {
+            if (m_spellInfo->Id == 1680)                   //Whirlwind
+            {
+                if (m_caster->GetAura(12292, EFFECT_INDEX_0)) //Sweeping Strike
+                    {
 
-			//This handles removal of one stack and makes sure it doesn't remove a stack if whirlwind doesn't hit anything.
-			SpellAuraProcResult unit(Unit *pVictim, uint32 damage, Aura* triggeredByAura, SpellEntry const * procSpell, uint32 procFlag, uint32 procEx, uint32 cooldown);
-					{
-					Aura* sweepingStrikeAura = m_caster->GetAura(12292, EFFECT_INDEX_0);
-					SpellAuraHolder* sweepingStrikeHolder = sweepingStrikeAura->GetHolder();
-					int32 CurrentCharges = sweepingStrikeHolder->GetAuraCharges();
+                    //This handles removal of one stack and makes sure it doesn't remove a stack if whirlwind doesn't hit anything.
+                        Aura* sweepingStrikeAura = m_caster->GetAura(12292, EFFECT_INDEX_0);
+                        SpellAuraHolder* sweepingStrikeHolder = sweepingStrikeAura->GetHolder();
+                        int32 CurrentCharges = sweepingStrikeHolder->GetAuraCharges();
 
-					Unit* pVictim;
-					Unit* target = pVictim;
-					//Adjusting range for the range of whirlwind to remove a charge (not litterly the range of whirlwind, just for how far the target can be away to drop a charge).
-					target = m_caster->SelectRandomUnfriendlyTarget(pVictim,8);
+                        Unit* pVictim;
+                        Unit* target = pVictim;
 
-					if (target)
-					{
-						sweepingStrikeHolder->SetAuraCharges(CurrentCharges - 1);
-					}
+                        //Adjusting range for the range of whirlwind to remove a charge (not litterly the range of whirlwind, just for how far the target can be away to drop a charge).
+                        target = m_caster->SelectRandomUnfriendlyTarget(pVictim, 8);
 
-
-					}
-				}
-		break;
-		}
+                        if (target)
+                        {
+                            sweepingStrikeHolder->SetAuraCharges(CurrentCharges - 1);
+                        }
 
 
-		if (m_spellInfo->Id == 20569 || m_spellInfo->Id == 845 || m_spellInfo->Id == 7369 || m_spellInfo->Id == 11608 || m_spellInfo->Id == 11609)  //Cleave all ranks
-		{
-			if (m_caster->GetAura(12292, EFFECT_INDEX_0)) //Sweeping Strike
-			{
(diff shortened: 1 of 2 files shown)


AndersSahlin:
I'm guessing you are making these changes with the Github web editor which might limit the usefulness of my reply, but still: proper commit messages are always helpful! ☺️

https://alistapart.com/article/the-art-of-the-commit

StadenElysium:
Please use spaces rather than tabs, since that's in keeping with the rest of the codebase.

src/game/Spells/Spell.cpp
```diff
@@ -3159,8 +3159,55 @@ void Spell::cast(bool skipCheck)
...
+				{
+
+			//This handles removal of one stack and makes sure it doesn't remove a stack if whirlwind doesn't hit anything.
+			SpellAuraProcResult unit(Unit *pVictim, uint32 damage, Aura* triggeredByAura, SpellEntry const * procSpell, uint32 procFlag, uint32 procEx, uint32 cooldown);
+					{
+					Aura* sweepingStrikeAura = m_caster->GetAura(12292, EFFECT_INDEX_0);
+					SpellAuraHolder* sweepingStrikeHolder = sweepingStrikeAura->GetHolder();
+					int32 CurrentCharges = sweepingStrikeHolder->GetAuraCharges();
+
+					Unit* pVictim;
```
FractalCabush: Are you sure this code compiles? Line 3170, looks like a function declaration because of the ";" in the end of that line, then you open a block as if it were a function definition. Maybe you intended to call the function?

Also, you create Unit* target and store the pointer of pVictim, but right after you assign it on line 3179. You already have the pVictim variable in-scope so redeclaring it on line 3176 is likely to "shadow" the variable.

Could it also be possible that pVictim is null?

if you want to declare Unit* target for use later you can just initialize it as Unit* target{0} or Unit* target = nullptr and then its safe and initialized to be used later in the code.

src/game/UnitAuraProcHandler.cpp
```diff
@@ -444,10 +444,22 @@ SpellAuraProcResult Unit::HandleDummyAuraProc(Unit *pVictim, uint32 damage, Aura
...
+			/*
+			Please note that UnitAuraProcHandler is not made to handle interactions like sweeping strike + Whirlwind/Cleave at all
+			reason being that UnitAuraProcHandler process the case for each hit and not per ability which results in a charge dropped per hit.
+			Instead modification to Spell.cpp has been made for whirlwind (SPELLFAMILY_WARRIOR) and cleave which handles the charge withdrawlas.
+			1 Charge is always withdrawn everytime you use whirlwind from spell.cpp which is why you will see +1 add charges in the prochandler here.
+			*/
+
+			//Fetch sweeping strike aura for being able to get and modify the amount the charges.
+			Aura* sweepingStrikeAura = GetAura(12292, EFFECT_INDEX_0);
+			SpellAuraHolder* sweepingStrikeHolder = sweepingStrikeAura->GetHolder();
```
FractalCabush: The typename already contains "Holder" and hints to the reader what its type and purpose is :) Adding an extra "Holder" in the variable name is added redundancy that removes readability, and the "Dont repeat yourself" principle ^^ just a tip!

src/game/UnitAuraProcHandler.cpp
```diff
@@ -478,6 +510,30 @@ SpellAuraProcResult Unit::HandleDummyAuraProc(Unit *pVictim, uint32 damage, Aura
 					triggered_spell_id = 12723;	//Note this SS id deals 1 damage by itself (Cannot crit)
 				}
 			}
+
+			// Case for Whirlwind and cleave. This will only run when procced by Whirlwind or cleave.
+			else if ((procSpell && procSpell->Id == 1680) || (procSpell && procSpell->Id == 20569) || (procSpell && procSpell->Id == 845) || (procSpell && procSpell->Id == 7369) || (procSpell && procSpell->Id == 11608) || (procSpell && procSpell->Id == 11609))
+
```
Cgbd42: You can check procSpell one time only

src/game/UnitAuraProcHandler.cpp
```diff
@@ -486,16 +542,56 @@ SpellAuraProcResult Unit::HandleDummyAuraProc(Unit *pVictim, uint32 damage, Aura
...
+					sweepingStrikeHolder->SetAuraCharges(CurrentCharges - 1);
+				}
+				else
+				{
+					target = pVictim;
+					triggered_spell_id = 22858;
+				}
+
+			}
+			triggered_spell_id = 22858;
```
Cgbd42: same line as 586, i guess one of theses is not supposed to be here

Tenarie: Wow, thanks for all the feedback ^^. I'm having a programming free day today (my eyes hurt).
But yea the code compiles fine and works as intended. I'll try do what you tell me but i don't really understand that many programming terms : P

src/game/Spells/Spell.cpp
Tenarie: Oh this was just for testing : ), forgot to remove it. I'm currently just doing try n' error to see what happens : ). 
Has been removed now.

src/game/UnitAuraProcHandler.cpp
Tenarie: Renamed variable : ).

src/game/UnitAuraProcHandler.cpp
Tenarie: Not sure what you mean here. It's just a huge OR statement where i check for whirlwind and all cleave ranks.

Are you saying i should make an if for each rank and use the same code ?

src/game/UnitAuraProcHandler.cpp
Tenarie: This is here since retaliation will not proc on single target if sweeping strike is up and no randomtarget will be near.
So if a randomunfriendlytarget can't be found then the target becomes the primary target and then procs the retaliation.

If i understand you correctly : ).

src/game/UnitAuraProcHandler.cpp
AndersSahlin: You can use boolean algebra to extract the `procSpell` check like this:

`if(procSpell && (procSpell->Id == 1234 || procSpell->Id == 5678 .... ))`

I would also consider replacing the hard coded numbers with variables or constants to improve code legibility and make it easier to maintain.

Tenarie: Ah yea, makes sense ^^

FractalCabush:
You are right! Didnt fully read your comments that stated that you are not yet versed in C++ :D if it compiles im happy!

What in mean is that you already have access to pVictim and you dont need to declare it again and you assign pVictim to target but then right after you assign it to something else :)

Skickat från min iPhone

src/game/UnitAuraProcHandler.cpp
Tenarie: It has been done : ), converted ID's to Variable name and if statement has been changed.

AndersSahlin:
If you're interested in some (in my opinion) best practice tips on programming in general I can recommend the book Clean Code by Uncle Bob:

https://www.amazon.com/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882

A bit old but still holds up for the most part...

m-laniakea:
Closing because this is superceded by two other commits.

PR #363, opened 2017-04-10
Fix for Touch of Zanzil

brotalnia:
Currently when opening the chest for the quest Klaven's Tower the Touch of Zanzil debuff doesn't get applied to the player, and if you apply the debuff to yourself using a gm command it doesn't prevent stealth. This is because the spell has the wrong effectAura2 and effectMiscValue2, and is considered to be applying stealth to the target, instead of making stealth unusable. So when the player opens the chest, the debuff doesn't even get applied because you can't be stealthed while looting or interacting with gameobjects.

There are several bug reports on the bugtracker about it that i can't read, but i suspect they are reporting this bug. https://elysium-project.org/bugtracker/search?title=zanzil&milestone=


QAkvastskaft:
I have made the tickets available to check if there is any information there that is helpful.

brotalnia:
It's the same issue. Their debuff was disappearing because the spell was considered to be applying stealth, and stealth breaks on damage. It should be working properly now.

PR #402, opened 2017-04-16
Miscellaneous World Database Fixes 2

brotalnia:
I've spent a lot of time going through the database errors and trying to correct each one. There are still a lot left, but this fixes a fair number of them. The dream is to eventually have a clean console on startup.


Kittnz:
Well done!

PR #403, opened 2017-04-16
Uses a proper ACE_HANDLE type for logging

martinjlowm:
Changes socket handle type to platform specific ACE_HANDLE instead of
explicitly casting the pointer to uint64.


martinjlowm:
I haven't tested this on Windows, please confirm that it actually compiles.

Issue #407, opened 2017-04-16
More skeletons missing immunitites: CONFIRMED


QAkvastskaft:
The following skeleton type mobs are missing the correct immune mask: 

MechanicImmuneMask = 213521

List:

- Commander Felstrom id: 771
- Rattlecage Soldier id: 1520
- Darkeye Bonecaster id: 1522
- Ravenclaw Champion id: 1869
- Hand of Ravenclaw id: 1870
- Eliza's Guard id: 1871
- Ravenclaw Guardian id: 1973
- Ravenclaw Regent id: 2283
- Puppet of Helcular id: 2946
- Splinterbone centurion id: 7346
- Dark Caster id: 8526
- Risen guard id: 10489
- Lord Darkscythe id: 10826
- Cauldron Lord Malvinious id: 11077
- Conquered Soul of the Blightcaller id: 12208
- Cauldron Lord Razarch id: 11076 (updated)

shmigy429:
> Cauldron Lord Razarch id: 13197

I'm seeing him as 11076, PR should be up soon.

QAkvastskaft:
Fixed: https://forum.elysium-project.org/topic/47461-development-update-24052017/

Issue #393, opened 2017-04-15
Cursed Mage bugged: CONFIRMED


QAkvastskaft:
Comment from QAkvastskaft:
This mob should have the same immunities as all other skeletons.

BUGGED BEHAVIOR
Cursed Mages are skeletons, like other skeletons they should be immune to incapacitate effects, stun and/or poison.

 CORRECT BEHAVIOR
The player shouldn't be able to incapacitate, stun and/or poison the Cursed Mages.

 SOURCES
Common knowledge

http://db.vanillagaming.org/?npc=8524

Original post on bug tracker here: https://elysium-project.org/bugtracker/issue/1863

QAkvastskaft:
Fixed: https://forum.elysium-project.org/topic/47461-development-update-24052017/

PR #412, opened 2017-04-17
Fix for Issue # 393/407, update more skeleton immunities

shmigy429:
Update Immune mask


shmigy429:
Should fix #407 and #393

Issue #394, opened 2017-04-15
Jordan's Weapon Notes: CONFIRMED


QAkvastskaft:
BUGGED BEHAVIOR
In the stated item, Jordan's Weapon Notes (id=6996), pages 3 and 4 are the same, meaning page 4 is a copy of page 3 and therefore is unneeded.

 CORRECT BEHAVIOR
Should be only once there.

 SOURCES
http://wowwiki.wikia.com/wiki/Jordan's_Weapon_Notes

Original post on bug tracker here: https://elysium-project.org/bugtracker/issue/1849

QAkvastskaft:
Fixed: https://forum.elysium-project.org/topic/47461-development-update-24052017/

PR #414, opened 2017-04-17
Fix Issue #394, correct page ordering

shmigy429:
Update order, delete duplicate


shmigy429:
Should fix issue #394, (hopefully removed references to the skeleton fix...)

brotalnia:
You created a second empty migration.

shmigy429:
I think that removed the duplicate empty migration

PR #419, opened 2017-04-17
Scepter Blue: Various Fixes

m-laniakea:
Everything should be fine with the Blue line.

Corrected quest issues.
Modified Number Two's immunities.
Added Narain's Letter.

Awaiting testing for Maws.

### DIFF
--- a/src/scripts/world/boss_anachronos.cpp
+++ b/src/scripts/world/boss_anachronos.cpp
@@ -20,12 +20,16 @@ EndScriptData */
 enum
 {
     NPC_ANACHRONOS          = 15192,
+    NPC_NARAIN              = 11811,
 
     // Anachronos's spells
     // -------------------
 
     SPELL_DIVINESHIELD      = 1020,
     SPELL_HOMESTONE         = 8690,
+
+    QUEST_WRATH_OF_NEPTULON = 8729,
+    LANG_NARAIN_LETTER      = 667,
 };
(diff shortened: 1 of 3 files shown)


src/scripts/world/boss_anachronos.cpp
```diff
@@ -95,6 +99,28 @@ struct boss_anachronosAI : public ScriptedAI
...
+    if (quest->GetQuestId() != QUEST_WRATH_OF_NEPTULON)
+        return false;
+
+    if (npc->GetEntry() != NPC_ANACHRONOS)
+        return false;
+
+    if (!player)
+        return false;
+
+    std::string letterText = player->GetSession()->GetMangosString(LANG_NARAIN_LETTER);
```
Cgbd42: I think you should use script_text here

m-laniakea: How come? You could be right, but I'd like to know why.

gheyhardy:
lmao hurry up unpaid dimwits they have a deadline to meet

src/scripts/world/boss_anachronos.cpp
Cgbd42: Mangos string is more for core utils, like innkeeper gossips, bg's strings etc..
Here you are on script side (src/scripts), so imo script_text is more appropriated.

PR #424, opened 2017-04-18
Fix object pools respawn after server restart

sipertruk:
Make object pools initialization happen at map population. Make object pools initialization take in account existing respawn time if no spawn is found, and then spawn the object at the right place.

Fix 2 problems :
-object pools init happens before respawn times are loaded from the db
-object pools init doesn't check if respawn times exist already 
Which  cause every object pools to respawn fresh at server restart when a different location is rolled. This include all sort of object including dungeon creature, even some boss.


StadenElysium:
Reverted.

PR #434, opened 2017-04-18
Re position of Gromsblood (BL)

Tenarie:
Repositioned the infamous underground gromsblood at the dark portal.


Jaxors:
This troublesome spawn has been around since pre-shutdown and it's high time that herb was re-planted. If you put it in the patch notes Herbalists will elect Tenarie their new king!

shmigy429:
I definitely wasted a few minutes trying to get that one...

brotalnia:
Why not update the position, instead of deleting the entry and then inserting a new one with a different guid. It won't be part of the pool this way.

m-laniakea:
Spawntime 25 seconds instead of the usual 45 minutes?

brotalnia:
Must be using Tenarie's special fertilizer :+1:

Tenarie:
haha, i asked in the discord and they said that it didn't matter since the herb spawn rate ´will over rule it.

Tenarie:
There you go sweeties ^^.

Issue #392, opened 2017-04-15
Watcher Callahan has Children's week quest: CONFIRMED


QAkvastskaft:
BUGGED BEHAVIOR
Watcher Callahan in Duskwood has a quest named "A Warden of the Alliance", which is a Children's week quest.

 CORRECT BEHAVIOR
The NPC originally acted as a cosmetic character or, as portrayed by many, Kzixx's bodyguard with no quests whatsoever.

 SOURCES
http://www.wowhead.com/npc=2142/watcher-callahan#comments

Original post on bug tracker here: https://elysium-project.org/bugtracker/issue/2441

m-laniakea:
He shouldn't give the children's quest, but he is not purely cosmetic. http://web.archive.org/web/20050209090847/http://wow.allakhazam.com:80/db/quest.html?wquest=236
http://web.archive.org/web/20050102153037/http://wow.allakhazam.com:80/db/quest.html?wquest=171

PR #442, opened 2017-04-19
Fix for #392 Remove children's week quest

shmigy429:
Remove his quest relation


shmigy429:
Should fix #392 Watcher Callahan has Children's week quest: CONFIRMED

PR #446, opened 2017-04-20
Fixing incorrect Silithid Carapace drop

BBCnewstoday:
Need to fix data at head of code...


StadenElysium:
Updated in another patch that will be added shortly.

PR #452, opened 2017-04-20
Fix dungeons reset time after server restart

sipertruk:
-Better loading of dungeons reset time after a server restart. And better saving too. On server restart the reset time was set at 0 (47 years ins the past)
-Also wipe the data field in the instance table along with creature respawn time in case of a reset happening after players exit (reset time in the past)

These 2 problems together can make players believe the instance has reset after server restart, all mobs respawn but the data field stays, which can block the script.

A little code clearing along the way.


brotalnia:
I am not getting saved to raids with this. I kill a boss, restart the server, and everything is respawned again including the bosses.

PR #455, opened 2017-04-21
fix for sql name of file (forgot to check the specs)

Jonathanfy:
I forgot the name format for the sql file.


Jonathanfy:
it's the same file as the old arcanite dragonling sql, just with the correct name.

brotalnia:
You forgot this part:
`INSERT INTO `migrations` VALUES ('20170407111859');`

Jonathanfy:
ty, i'm not completely used how you deal with database changes yet.

Issue #271, opened 2017-04-04
[Quest] A Hero's Reward


Xantium995:
Currently, the quest A Hero's Reward (id 7486) is not obtainable after completing Hands of the Enemy (id 6824), which is I believe the only requirement to unlock the last quest.
This quest, given by Duke Hydraxis, has been added to the game as part of patch 1.3.
I haven't found any reputation requirements for this quest, so I suppose it should be pickable with the Honored reputation level.

http://db.vanillagaming.org/?quest=7486
http://wow.gamepedia.com/Quest:A_Hero%27s_Reward

QAkvastskaft:
Fixed here: https://forum.elysium-project.org/topic/47461-development-update-24052017/

PR #457, opened 2017-04-21
Fix for [Quest] A Hero's Reward #271

shmigy429:
Allow quest to be picked up after the chain is complete.


shmigy429:
Should fix #271

thepparker:
Changes like these should be done in the itemization patches rather than core migrations, as they are dependent on patch level (the quest was introduced with 1.3).

PR #459, opened 2017-04-21
Creature dmg Multiplier fix + items proc issue fix.

BBCnewstoday:
Noticed that some creatures dmg are modified to huge numbers(eg NPC 15554 is x4 dmg and hitting harder than chromaggus)

And old item-proc issues as seen in bug reports:
https://elysium-project.org/bugtracker/issue/3649 - Thunderfury
https://elysium-project.org/bugtracker/issue/4073 - other items


brotalnia:
The file has no extension.

BBCnewstoday:
Added extension, brotalnia thank you! =)

m-laniakea:
The proc rate for this item is not constant throughout all Patches and thus belongs over in /itemization.

PlaskED:
They actually replied to that bugreport and applied the fix to the main server. It was however reverted the next week. Unfortunately Thunderfury will be gimp just like it was in 1.9.

BBCnewstoday:
Added to patch issue with custom buffed NPC.

StadenElysium:
The proc rate for Thunderfury is incorrect (it shouldn't be the same across all patch levels) and if my understanding is correct, some of the multipliers are increased with reason and applying would cause regressions. The data for the proc rate on the other weapons isn't good enough to justify the buff - it should be opened as an issue of its own for further research.

Issue #490, opened 2017-04-24
Arcanite Dragongling broken 


StadenElysium:
Apparently damages itself with its own Flame Buffet.

Fyllhund:
xD, i'll fix it.

Jonathanfy:
xD, I'll fix it when I get home.

Teadrinker1234:
It is happening on darrowshire too. Please fix!!!

strudelkuchen:
Still broken.

QAkvastskaft:
Fixed here: https://forum.elysium-project.org/topic/47461-development-update-24052017/

PR #494, opened 2017-04-24
Arcanite dragonling attack other (#490)

Jonathanfy:
Y, now the arcanite dragonling should attack someone else.

Fix for issue #490


Teadrinker1234:
when is this getting fixed so it doesn`t flame buffet itself?

strudelkuchen:
> when is this getting fixed so it doesn`t flame buffet itself?

+1

Issue #62, opened 2017-03-19
Defender of the timbermaw (exalted trinket) NOT WORKING


lequintal69:
defender of the timbermaw http://db.vanillagaming.org/?item=21326#wh-comments is completly bugged right now.

**CURRENT BEHAVIOUR:**

-summoned minion just follows you and almost never attack (90% of the time he does nothing).
-Only twice out of 20 summons, he casted ONE single lightning bolt doing 500dmg during its entire duration, then went back to bugged state
-never heals
-has 427 hp

**CORRECT BEHAVIOR**

-should have around 500hp
-should do around 220-300 damage (dmg seen in videos ranges from 233 to 286). Can crit
-should heal when you are below 50% hp (estimated values from comments and videos).
-heal values is around 500-600 (can crit for 800-900) as seen in the video
-casting time is around 2sec
-will melee if target is immuned to nature dmg
-doesnt break CC (sheep,trap, etc)

nb: healing is top priority, meaning if you (or your pet, or someone else in your group) drop below 50% health, he should start healing until you are above 50%

**SOURCES**

http://db.vanillagaming.org/?item=21326#wh-comments comments
https://youtu.be/42QFS7qJyVo?t=5m35s (dmg, heal, crit heal, casting time)
https://youtu.be/4c4LjRjUMMs?t=22m31s (dmg, heal, casting time)
https://web.archive.org/web/20061231193841/http://www.thottbot.com:80/?i=52390  (many many comments on behaviour, check last one)

lequintal69:
https://web.archive.org/web/20061231193841/http://www.thottbot.com:80/?i=52390

many usueful comments here aswell.
Last one is quite precise
"Average heal: (about 2 sec cast) 513hp 
Average lighting  bolt damage:  (about 2 sec cast)  264  (It seems to do more damage now, post 2.0)  
Average melee damage: 85 (against targets with essentially no armor) "

lequintal69:
@Jonathanfy  I have seen your dragongling and cthun trinket fixes, any chance on fix for this trinket ? Would make a lot of pve and pvp tryhards happy :)

Jonathanfy:
Y, i'll look at it

QAkvastskaft:
Fixed here: https://forum.elysium-project.org/topic/47461-development-update-24052017/

PR #528, opened 2017-04-27
vanquished tentacle fix + timbermaw ancestor 

Jonathanfy:
vanquished tentacle fix.

Increased its' number to hit between 90 and 140. 

https://elysium-project.org/bugtracker/issue/4259

Had fuck-all to go by though.

Also threw in ai for the timbermaw ancestor trinket.  #62


GuybrushGit:
Missing semicolons at the end of the lines.

Also the numbers should be higher:
https://www.youtube.com/watch?v=OGZnQh4Vth8
220 / 500 crit on a low level

https://www.youtube.com/watch?v=wEcPkwsfLd4&feature=youtu.be&t=37s
150 on a warrior

Should be like min 200 max 250.

Jonathanfy:
watya on about. U don't need to close an sql row with semicolons :) but sure, i'll update it

brotalnia:
You do need a semicolon if there is more than one line. Otherwise it will give an error after the first one.

Jonathanfy:
something particular about mysql?

Jonathanfy:
fuck :D added the fix for timermaw ancestor aswell

Seagulls:
Typos in  src/scripts/world/npcs_special.cpp

+struct npc_TimermawAncestorAI : ScriptedPetAI

should be +struct npc_Tim**b**ermawAncestorAI : ScriptedPetAI etc

sql/migrations/20170430191920_world.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO migrations VALUES ('20170430191920');
+UPDATE creature_template SET scriptName = 'npc_TimermawAncestor' WHERE entry = 15720;
```
brotalnia: You forgot to change the name here. It says Timermaw instead of Timbermaw.

src/scripts/world/npcs_special.cpp
```diff
@@ -1479,6 +1479,84 @@ CreatureAI* GetAI_npc_arcanite_dragonling_dragonling(Creature* pCreature)
...
     return new npc_arcanite_dragonling_dragonlingAI(pCreature);
 }
+
+/*######
+## Timbermaw Ancestor
+######*/
+enum
+{
+	SPELL_HEALING_TOUCH = 11431,
+	SPELL_LIGHTNING_BOLT = 6041
```
brotalnia: According to wowhead the correct spell id for the lightning bolt is 9532.

http://www.wowhead.com/npc=15720/timbermaw-ancestor#abilities

brotalnia: Also you are using tabs instead of 4 spaces.

GuybrushGit:
Damn did you do any research or are you just making shit up?
Lightning bolt 220 - 280 and no cooldown, most likely 10391.
Healing for 200-500, most likely 5187, where did you get the 10 minute cooldown?
Health should be 500.
Also script name doesn't follow convention and you're still using tabs instead of spaces.

Did you even test ingame before submitting?

Jonathanfy:
all spells were from the reporter. If anything is wrong I am more than willing to change something. About spaces, y I forget sometimes sorry.

brotalnia:
@GuybrushGit This lightning bolt is a player spell, mobs usually have their own spells. The spell from wowhead exist in all expansions which makes sense because its used by the furbolg, and it's damage scales with level. It does exactly the correct amount of damage when used by the furbolg. Should be pretty obvious that's correct spell.

GuybrushGit:
I see. Both spells casts are still wrong though, they're being cast as triggered instead of casted, the timers are wrong, the heal shouldnt be a custom for 500.

lequintal69:
@GuybrushGit 

I asked Jonathanfy to kindly fix it, but we wasnt very clear on some aspects of the trinket:) I'm gratful he is giving it a go

To sum up: (sources are in original bug report):

1- pet should be defensive (will stop attacking if target is  sheeped, trapped, sapped, etc)
2- lasts 30s
3- defaut behavior is casting lightning bolt. Would only ever melee if target is immuned to nature dmg, or if silenced or oom (melee does around 85dmg to low armor targets)
4- minion remains when you swap trinket
5-  Should heal when you are below 50% hp and keep healing until you go above 50%
6- only has 500hp

nb: check behaviour in the videos in posted
https://youtu.be/42QFS7qJyVo?t=5m35s   Heals RIGHT away when summoned because summoner has less than 50% hp then LB
https://youtu.be/4c4LjRjUMMs?t=22m31s   Heals RIGHT away (same reason) then spams LB
thotbot comment:   _he'll easily throw out 7 or 8 heals for about 500, and if nobody needs it, chain nuke for 200-300_


about spells
thotbot comment talks about 2sec duration for both LB and healing touch
_Average heal: (about 2 sec cast) 513hp 
Average lighting  bolt damage:  (about 2 sec cast)  264_

from the videos i posted, you can see they have the same casting duration. As to know if its 2s as the comment suggests, 2.5 or even 3s, i'm not sure. Could very well be 3s.

I think those are the right spells
healing touch : http://db.vanillagaming.org/?spell=23381   (bottom heal:  448)
lightning bolt:   http://db.vanillagaming.org/?spell=10391  player's lightning bolt 227 to 256dmg
Couldnt not find another one that matched the values from video and thotbot comments

EinBaum:
Some style errors: 
1. Put a line break at the end of the file
2. Don't use tabs. Spaces only.

Jonathanfy:
Ty for the input. The only thing I Don't have working is to stop casting lightning bolt if target is immune.

lequintal69:
Tbh, the immune nature thing is really uncommun anyways. Will never happen in pvp, and in pve, if any boss is immuned (i cant think of any except viscidius), you are not going to use this trinket for the low melee dmg anyways:p

thepparker:
The second video shows only one heal and lightning bolt spam even with the caster on <50% hp, there must be some cooldown on the heal or other AI at play (such as only healing if there's nothing to lightning bolt - http://db.vanillagaming.org/?item=21326#comments:id=324044). He should also pick party and raid members to heal

Jonathanfy:
y, that was what I thought. There has to be some sort of cooldown on the heal otherwise it's op as fuck.

lequintal69:
There might be a low cooldown on healing touch then. 

Annnnd Here is what i found on thotbot mirrors:

_Casts heals on yourself and group members under 50% hp **refresh time on heal is 3 seconds**. casting time on heals and nuke is 3 seconds._ 

and from another guy

_if your hp is below 50% and heals you ever **5** seconds for about 460-500 hp_

So, its something between 3 to 5seconds cooldown.
That would match with the 2nd video i posted where the minion instantly heals when summoned, then does 2 lightning bolts then heals again.

That would make it slightly less good but still really decent.

Dont forget that it has 500hp only. A single Boss AOE, a single fireblast, multishot, cone of cold, etc can kill it, so it is only good in very specific fights

m-laniakea:
Three things to fix now:
1) Spacing
2) Cooldown on heal (9s-10s)
3) Ancestor will currently only heal the owner. You can use `DoSelectLowestHpFriendly`

lequintal69:
it is not 9-10sec on healing
it is either 3 or 5

GuybrushGit:
Side note: there are a ton of other guardians that need to be scripted as well, I tested using EventAI and generic_spell_ai but they don't support player pets.
In the future we should implement cmangos [GuardianAI](https://github.com/cmangos/mangos-classic/blob/39ab48e42ad8beacbb3b695c7ce88daca8b5b1cb/src/game/AI/BaseAI/GuardianAI.cpp) rather than scripting these pets 1 by 1.

lequintal69:
@m-laniakea 

"AI cooldowns start at the begin of a cast, regardless of whether or not the cast successfully completed"
Didnt know that. 7sec makes sense then.

Anyone able to implement it ? its not much of coding, is it?

Jonathanfy:
"You can use DoSelectLowestHpFriendly" then we'd have to implement the guardian ai class which has that method. For now the timbermaw only heals it's owner on a 7 sec cooldown and casts lightningbolt.

PR #536, opened 2017-04-28
fix vmaps assembler build

futjikato:
add `game` and `game/Commands` to build directories ( `World.h` gets included in `VMapsManager2.cpp` and `World.h` includes some scripts from `game/Commands` )
add flag to compile with cpp 2011 flag

Please check if the compiler flag is correct this way. Im not the most savvy c++ dev so there might be a better option for this.


thepparker:
You can add new compiler options with the cmake command ADD_DEFINITIONS("-std=c++11") rather than setting the flags directly

futjikato:
@thepparker tanks for the info will remove that and rebase. What about the directory includes? is that a bug or am i missing something else?

thepparker:
There are a number of missing or bad include directories in almost every contrib, you'll have to fiddle with them to get them to compile.

Here's my diff list for compiling all extractors/assemblers: https://gist.github.com/thepparker/52f70f9fce90dda5527c2bb037635b12

I have a separate copy of the repo for the extractors.

futjikato:
Thanks a lot. Will adopt certain changes I need ( such as uncommend `ParsMapFiles()` call ... took me way too long to figure that one out )

So are PRs regarding contrib stuff not expected anyway? Seems kinda "meh" to have them partilly broken in this repo

brotalnia:
You guys know that you can compile all extractors using the Giperion branch, right?

thepparker:
Some of my additions are based on his and the stock mangos branches, though he added a few extra libs that weren't necessary so I just kept it minimal - whatever suits

futjikato:
Gonna close this PR. Dont think this helps anyone ;) Thanks again for your tipps. Got my docker build running now

Issue #535, opened 2017-04-28
BG Marks last longer than 24 hours in mailbox.


Slicy:
Bugged Behavior
Marks are lasting longer then 24 hours in Mailbox

 Correct Behavior
Marks in mailbox should expire after 24 hours

Sources

**http://beust.com/wow-battlegrounds.html** 
_"Be aware that each mark only stacks up to 20, so make sure you never have more than 17 marks when you enter a battleground, or you might end up losing some (the extra marks are actually mailed to you in that case, but they have a twenty four hour expiration, so check your mailbox as soon as possible to retrieve them or they will be lost for good)."_

Date is August 8th 2006 so 5 months or so before TBC Pre-patch which wouldn't be a patch we are at currently but all patch notes from changes to mailbox use listed below.

http://vanilla-wow.wikia.com/wiki/Mail

Shows all the patch changes made to mail box use.

Priority should be extremely high as this is getting abused by a handful of players rotating BG Marks stacking in Mailbox beyond Mailbox caps in order to "markbomb" 10 minutes before each reset and screw up the actual rankers.

This bug has made ranking a complete shitfest where legit players can't farm beyond what others with malicious intent are getting just before restart by abusing this major bug, since weeks/months.
It is killing the ranking scene as the players ranking up have no incentive to play the  game legitly and farm their honor normally since they are getting screwed over by bug abusers, lose 1-2 weeks and have to eventually give up no matter how hard they farmed honor and played the game.

AndersSahlin:
So... They seem to have "expired" after 3 days according to the live code, not 4-5 weeks as the bug report states. 

Isn't the actual issue that the marks can be used even after the mail has expired? 

https://elysium-project.org/bugtracker/issue/4014

GuybrushGit:
Is there any trick to reproduce that? The emails get deleted properly when I relog.

Slicy:
Not supposed to be 3 days nor 30 days. 24 hours is the maximum duration of BG marks once you receive them in mailboxes in vanilla according to this source.

Lorilay2:
Expired mails just sit in your inbox with nothing listed where the expiration time would normally be.

AndersSahlin:
Possibly fixed by #734.

PR #538, opened 2017-04-28
Lower BG mail expire time to 1 day

GuybrushGit:
Related to https://github.com/elysium-project/server/issues/535


thepparker:
I think the issue is that expired mails are not being deleted correctly, if at all. I've noticed the same with expired mail not being returned on my bank alts. I had a brief look and couldn't see anything glaringly obvious that would cause it, but perhaps I missed something or something has been changed server-side that we are not privy to.

GuybrushGit:
I don't see the issue as well, unless people are staying online 24/7 to not have them deleted.

Issue #295, opened 2017-04-06
Warlock Summon spell targeting does not work correctly.


kevmodrome:
Switching target during the process of summoning someone changes the summoning target. The summon spell should summon the current target when the summoning spell is actually started. Switching target after should not matter at all.

aeranadots:
Have also noticed this, the summoning animation is also not working for me. When players click the summoning portal, they should start channeling until enough players have clicked it, iirc.

kevmodrome:
Yep! That's not as irritating as not being able to switch targets though :)

kakysha:
This bug becomes a bigger problem when you have some addons that inspect players on mouseover over players, for example to add additional info into tooltip. Because inspecting internally causing target change, so when you summoning target and hover your cursor over another players around you - your target is changed and summon is fked up. It becomes almost impossible to summon anyone if you are in raid with 40 ppl around you :)

GuybrushGit:
I could not find the spell that triggers the channel animation for the players that click it. Anyone knows what is it?

sipertruk:
Maybe it's "arcane channeling" 23017.

QAkvastskaft:
Fixed here: https://forum.elysium-project.org/topic/47461-development-update-24052017/

PR #541, opened 2017-04-28
Fixes to Ritual of Summoning.

GuybrushGit:
https://github.com/elysium-project/server/issues/295 **Warlock Summon spell targeting does not work correctly**

Summons now use the target the lock has when the channeling starts, not the one when the channeling completes. Had to store the target in the GameObject the ritual creates.
Added the animation spell for the clickers as well.

This may also solve the issue where people summoned while flying keep flying after the teleport.
The unit target used in the previous system didn't always have the UNIT_STAT_TAXI_FLIGHT flag set.


thepparker:
Very nice. If we want to get even more blizzlike, when one of the users who clicks the summon cancels the channel (i.e. moves), the summon should also be cancelled. At the very least they should probably be removed from the users list on cancel.

Issue #539, opened 2017-04-28
Golemagg's 2 adds (Core Rager) have no Threat table for Paladins


theloras:
**BUGGED BEHAVIOR**
Core Rager's threat table does not include Paladins/Holy Damage. This is a huge bug that prevents paladins from being able to tank MC40.

As it stands right now, they respond only to Taunt and produce no threat whatsoever in response to any damage or healing.

**CORRECT BEHAVIOR**
Core Rager's should acknowledge Paladin's threat table for tanking.

**SOURCES**
http://wowwiki.wikia.com/wiki/Threat
http://wow.gamepedia.com/Core_Rager
http://db.vanillagaming.org/?npc=11672#wh-comments

Original Bug Tracker post - https://elysium-project.org/bugtracker/issue/3136

QAkvastskaft:
Fixed here: https://forum.elysium-project.org/topic/47461-development-update-24052017/

PR #544, opened 2017-04-28
Fix to Core Rager AI

GuybrushGit:
https://github.com/elysium-project/server/issues/539 **Golemagg's 2 adds (Core Rager) have no Threat table for Paladins**

Core Ragers weren't checking their threat table for targets.


theloras:
@GuybrushGit - you sir are simply amazing :)

Issue #534, opened 2017-04-28
Ignite currently does not get partially resisted - It should.


kevmodrome:
Hi! From what I understand from our mages in our guild Ignite does not get partial resists. This is not how it worked back in vanilla. For proof you can see this video: https://youtu.be/DtzbY-7tTVc?t=40s (should show up around 44-45 second mark). You find the information in the bottom right in the combat log. This bug trivializes some of the encounters in AQ40 quite severly. Please fix quickly! :)

GuybrushGit:
Ignite does get partially resisted but its following the same rule as other dots, which means it resists far less than a direct spell.
Some dots however have been proven to follow direct spell rules:

>// NOSTALRIUS: Some DoTs follow normal resist rules. Need to find which ones, why and how.
>// We have a video proof for the following ones.
>case 23461:     // Vaelastrasz's Flame Breath
>case 24818:     // Nightmare Dragon's Noxious Breath
>case 25812:     // Lord Kri's Toxic Volley

Maybe the devs should consider adding ignite to this list.

kevmodrome:
I see! I have no idea which one is correct, kind of hard to find old sources. I have found a blue post from 2010 (I know, not vanilla, but it's still before the ignite overhaul in later expansions) saying this: "While the means of arriving at the effect is different, Ignite will function similarly to Immolate, as any DoT component remains subject to this potential resist mechanic. ". Now, this doesn't really say anything except that it is similar to immolate and can get partially resisted unfortunately. I found some other information on wowwiki from 2008 that says "For DoT spells, the partial resistance calculation is applied on each tick.".

I would assume that it is handled the same way as Corruption and/or Immolate for warlocks.

sources:
- http://blue.mmo-champion.com/topic/73470-bug-ignite-still-double-penalized/
- http://wowwiki.wikia.com/wiki/Resistance?oldid=1401234

GuybrushGit:
>Previously, dots in general were 1/10th as likely to be resisted as normal spells. 
>http://web.archive.org/web/20080601184008/http://forums.worldofwarcraft.com/thread.html?topicId=65457765&pageNo=18&sid=1#348

kevmodrome:
Interesting! I wonder if they mean the application of the spell, the actual damage parts or both. To me it sounds a lot more probable that the damage is resistable as normal and the the application itself is not (i.e. 1/10 as likely to get resisted). Hard to know for sure regardless I guess.

QAkvastskaft:
[...]

PR #647, opened 2017-05-09
Fix to Ignite dot damage

GuybrushGit:
This fixes the caster and target modifiers from being applied to the ignite dot damage.

Another issue related to ignite is the partial resist rate, see my first comment here: https://github.com/elysium-project/server/issues/534
In the video posted there the ignite resists twice in a short amount of time, let me know if you consider this enough to make ignite follow normal resist rules and I'll add it to this PR as well.


thepparker:
The only issue I can see with this is that the ignite damage won't be increased by the stack amount anymore. We'll have to multiply it by the stacks manually on trigger. 

Something like 

`tickDamage = tickDamage / (2 * igniteAura->GetStackAmount());` when calculating the rollover damage
and then
`tickDamage *= igniteHolder->GetStackAmount();` after we mod the stack amount.

In testing ignite ticks seem to resist often enough, including 2-3 partial resists in a row, so I'm not sure about changing its classification.

GuybrushGit:
@thepparker I don't think I understand, my change only prevents outside modifiers from being applied. I don't see how it affects the damage being incremented on each tick here: 
https://github.com/elysium-project/server/blob/400cb2d64ed0c1c0598ce0867314d38e47fb2103/src/game/UnitAuraProcHandler.cpp#L738

thepparker:
Ignite damage not only rolls over, but because it is a stacking ability, the damage is multiplied by the stack which is handled in Unit::SpellDamageBonusDone (https://github.com/GuybrushGit/server/blob/0e73e401cba55c6131a1161eeb2ddd2470d72510/src/game/Objects/Unit.cpp#L6260). If we don't multiply the stack, ignite is only ever ticking for the rollover damage which usually ends up around 700.

thepparker:
Actually, it appears that the stack damage is handled in SpellAuraHolder::SetStackAmount. The stack handling in Bonus* is just for that, the bonus addition. With your changes I can't get ignites above around 1.5k with 1000 spellpower, so something is off. I think the stack application is broken since it is using the prototype base damage rather than our modified damage, and the bug was hidden due to other factors. I'm going to fiddle around with it some more.

GuybrushGit:
This is with my patch, using a spell that always does 750 (felfire): 
https://streamable.com/kxkix
Spamming a ton it caps out at 750 ticks (150*5): https://streamable.com/t6elx
This is the intended behaviour right?

The only possible issue I found is the initial damage being carried, even if we use a low damage spell, see:
https://streamable.com/1f7td

Unless this is intended as well?

thepparker:
The low damage crit keeping ignite up is intended, any fire crit will maintain the ignite. I agree, damage checks out. I was wrong about the stack multiplying, ignite is weird in that regard. Seems all the stack is there for is to cap the roll. 

Preventing ignite from double scaling with caster mods is correct, that will stop it growing out of control when building stacks. However, I'm not sold on preventing damage taken mods from double dipping. https://youtu.be/bMWKNeIISmo?t=97 clearly demonstrates the damage jumping when CoE is applied. Whether the mod is taken into account on later ignite applications to prevent them doubling up is unknown, but I think it is a feature of ignite.

If we let damage taken mods double dip (say, 5 stack imp scorch and CoE), we get a 1.265 increase on the ignite which checks out. Nothing obscene like caster mod scaling.

Other than that, I'm happy.

GuybrushGit:
Seems like a bug to double dip like that but its blizzlike, reverted that part.

m-laniakea:
@GuybrushGit Ignite mechanics were confirmed working by Nano, including Level 1 fireballs keeping the tick damage going. It's just the interaction with spell damage mods increasing the tick damage, which caused the problems. Ignite damage should depend only on the fire crit amount, and your two lines in Unit.cpp should take care of that.

PR #693, opened 2017-05-12
WorldSession::HandlePlayerLogin: redo pending teleport at login

WobLight:
If teleport semaphores are set, the server will wait for a reply from the client,
if the client just logged in, the teleport ACK was lost and no reply will be sent,
This will cause the server to ignore client movements until semaphores are cleared.

This fixes a common issue for players who get their corpse removed (by Remove Insignia)
by another player while disconnected in battlegrounds.


WobLight:
I was thinking about possibly adding a `function<void()>` parameter to `Player::TeleportTo`, allowing caller to store a specific function to restore the teleport when lost instead of guessing how to do it at player login. It would be far more flexible approach.

WobLight:
well then, following changes may be unexpected side effects somewhere else:
- FindPlayerNotInWorld was check for player actually to be in world, that has been corrected
- Players are not instantly kicked when crashing on a loading screen (teleported far)
  (e.g. a player crashing while loading a BG will relog in BG instead of getting deserter due to forced logout)

WobLight:
fixes https://elysium-project.org/bugtracker/issue/520

PR #734, opened 2017-05-15
Make mailed Honor Marks expire

m-laniakea:
Honor Marks could be stored in the Mail Box indefinitely


GuybrushGit:
Expiration time should be lowered to 1 day as well, currently its 3 days, see #535.

m-laniakea:
@GuybrushGit Yeah, changed it while you were posting

thepparker:
Is the same mask (MAIL_CHECK_MASK_NONE) applied to forwarded mails with items as well? I thought it was just a generic issue with expired mails not being returned/deleted.

PR #745, opened 2017-05-17
Uldaman altar event left broken after #541

sipertruk:
Prevent the interruption of the ritual when the number of uses is reach.


GuybrushGit:
My bad.
The proper fix would be to check if the ritual is persistant before stopping the channelling:

Add the part in bold to GameObject.cpp line 1523:
if (user && user != owner **&& !info->summoningRitual.ritualPersistent**)

PR #755, opened 2017-05-17
Replace auto_ptr and remove some manual memory management

StadenElysium:
Not finished, need to find better solutions for the queue

### DIFF
--- a/src/framework/Network/MangosSocketImpl.h
+++ b/src/framework/Network/MangosSocketImpl.h
@@ -7,7 +7,6 @@
 #include <ace/os_include/sys/os_socket.h>
 #include <ace/OS_NS_string.h>
 #include <ace/Reactor.h>
-#include <ace/Auto_Ptr.h>
 
 #include "MangosSocket.h"
 #include "Common.h"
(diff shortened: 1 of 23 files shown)


WobLight:
While I was comparing that with my branch, I've noticed that MassMail doesn't actually need a copy constructor, using emplace_back here should be enough https://github.com/elysium-project/server/blob/867c55b13e1ed32ecb1fd08f982df43a7f4543be/src/game/Mail/MassMailMgr.h#L92
Also, my build skips recompile after `MassMailMgr.h` is modified...

src/framework/Network/MangosSocketImpl.h
```diff
@@ -336,7 +332,7 @@ int MangosSocket<SessionType, SocketName, Crypt>::handle_input_payload(void)
     MANGOS_ASSERT(m_Header.space() == 0);
     MANGOS_ASSERT(m_RecvWPct != NULL);
```
WobLight: replace `NULL` with `nullptr` or use `!m_RecvWPct`
same for https://github.com/elysium-project/server/blob/867c55b13e1ed32ecb1fd08f982df43a7f4543be/src/framework/Network/MangosSocketImpl.h#L290
https://github.com/elysium-project/server/blob/867c55b13e1ed32ecb1fd08f982df43a7f4543be/src/framework/Network/MangosSocketImpl.h#L339 to be replaced with `reset()`

PR #759, opened 2017-05-18
Weapon skill fix

Cephel:
The existing melee/ranged miss calculation based on weapon skill shows some deficits that aren't in line with existing research, particular an erroneous "bump" of skill between two specific weapon skill values. They also completely devalue half the weapon skill items by providing zero benefit (+hit can only be obtained in full intervals), both of those phenomena were not present in vanilla WoW.

Shown below is the weapon skill effects on hit/miss chance shown side by side between new and old formula:

Clearly shown is the massive bump in hitrate when going from 304 to 305 weapon skill (effectively 2% hit) that has no business being there. The corrected formula smoothes the transition between large and small weapon skill formulas by adjusting the hitcap by 0.4%, resulting in a value of 8.60%, an often quoted figure with tons of research behind it, as opposed to the 9% currently implemented on the server. Further, the lower cap is raised by an equal amount to 5.60% and while this figure has less research, the figure of 5.60% is present **everywhere**, from defense skill crit calculations, to mob crit strike chances, etc. and as such a reasonable assumption can be made that the 0.6% difference in base miss chance is valid from the 5% currently active on the server.

To summarize, these would be the effects of implementing these changes:
- Scaling remains unchanged (0.1% at higher differences, 0.4% at lower)
- Scaling switch point (the point at which the weapon skill formulas switch over) remains unchanged
- Hitcap with 300 weapon skill is lowered from 9.00% to 8.60%, effectively no change
- Hitcap with 304 weapon skill (Maladath, Anubisath Warhammer, etc.) is lowered from 7.40% to 7.00%, effectively 1%
- Hitcap with 305 weapon skill (racials) is raised from 6.00% to 6.60%, effectively 1%
- Hitcap with 307 weapon skill (Edgemasters) is raised from 5.80% to 6.40%, effectively 1%
- Hitcap with 312 weapon skill (racial + Edgemasters) is raised from 5.30% to 5.90%, effectively no change, although it should be noted that with the corrected formula you gain an extra 1% hit over just the racial 305 skill

### DIFF
--- a/src/game/Objects/Unit.cpp
+++ b/src/game/Objects/Unit.cpp
@@ -2738,11 +2738,11 @@ float Unit::MeleeSpellMissChance(Unit *pVictim, WeaponAttackType attType, int32 
 
     // PvP - PvE melee chances
     if (pVictim->GetTypeId() == TYPEID_PLAYER)
-        hitChance = 95.0f + skillDiff * 0.04f;
+        hitChance = 94.4f + skillDiff * 0.04f;
     else if (skillDiff < -10)
-        hitChance = 93.0f + (skillDiff + 10) * 0.4f;        // 7% base chance to miss for big skill diff (%6 in 3.x)
+        hitChance = 93.4f + (skillDiff + 10) * 0.4f;        // 7% ~ 6.60% base chance to miss for big skill diff
     else
-        hitChance = 95.0f + skillDiff * 0.1f;
+        hitChance = 94.4f + skillDiff * 0.1f;
 
     // Hit chance depends from victim auras
     if (attType == RANGED_ATTACK)
@@ -3079,8 +3079,8 @@ float Unit::MeleeMissChanceCalc(const Unit *pVictim, WeaponAttackType attType) c
     if (!pVictim)
         return 0.0f;
 
-    // Base misschance 5%
-    float missChance = 5.0f;
+    // Base misschance 5.60%
+    float missChance = 5.60f;
 
     // DualWield - white damage has additional 19% miss penalty
     if (haveOffhandWeapon() && attType != RANGED_ATTACK)
@@ -3104,7 +3104,7 @@ float Unit::MeleeMissChanceCalc(const Unit *pVictim, WeaponAttackType attType) c
     if (pVictim->GetTypeId() == TYPEID_PLAYER)
         missChance -= skillDiff * 0.04f;
     else if (skillDiff < -10)
-        missChance -= (skillDiff + 10) * 0.4f - 2.0f;       // 7% base chance to miss for big skill diff (%6 in 3.x)
+        missChance -= (skillDiff + 10) * 0.4f + 1.0f;       // 7% ~ 6.60% base chance to miss for big skill diff
     else
         missChance -=  skillDiff * 0.1f;
 
--- a/src/game/Objects/Unit.cpp
+++ b/src/game/Objects/Unit.cpp
@@ -3104,7 +3104,7 @@ float Unit::MeleeMissChanceCalc(const Unit *pVictim, WeaponAttackType attType) c
     if (pVictim->GetTypeId() == TYPEID_PLAYER)
         missChance -= skillDiff * 0.04f;
     else if (skillDiff < -10)
-        missChance -= (skillDiff + 10) * 0.4f + 1.0f;       // 7% ~ 6.60% base chance to miss for big skill diff
+        missChance -= (skillDiff + 10) * 0.4f - 1.0f;       // 7% ~ 6.60% base chance to miss for big skill diff
     else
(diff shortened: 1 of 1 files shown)


GuybrushGit:
>Clearly shown is the massive bump in hitrate when going from 304 to 305 weapon skill (effectively 2% hit) that has no business being there.
Why not? Do you know for sure why Blizzard created the bump? It could have been intentional.

This "bump" carried over to TBC when the expertise system was introduced, for some reason they didn't change it.

Cephel:
The "bump" is based on TBC research and as you said was a consequence of changing the weapon skill / expertise system in TBC and does not apply to Vanilla.

gemt:
Interesting stuff, but you repeat "very well researched" a lot, but reference no sources. I know you have them, my point is simply, would it be possible to add some of them for completeness?

Cephel:
Assembling all the research again will take a bit of time, as it is extremely exhaustive. I will update this post as soon as able.

It should however be noted that the bump between 304 and 305 has absolutely zero credibility in and of itself and practically no research to even mention such a thing, so if the argument is that lack of enough conclusive research (there was never any confirmations during Vanilla on the topic from any official source either way) is enough to decline this patch, then atleast this 2% hit bump must also be removed, as it is way more "out there" than anything I've implemented here.

The patch can be broken down into several components, each should be addressed individually:
- The 0.60% offset, there is a tons of supporting evidence, but no direct proof, because a 0.6% difference is extremely unnoticable. However, the existence of this number in **literally every other melee mechanic** is very strong argument that it should be taken as correct. As an example, the hitcap for tanks is 300 + 140 = 440, 140 * 0.04% (crit immunity per point of defensive skill) is exactly 5.60%, this number is found throughout all relevant melee chance calculation, ranging from mob crit chances, dodge, parry, scaling, etc. 
- Removing the 2% gap between 304 and 305, as mentioned above there is no evidence that this should be here and Occams Razor is almost enough by itself to have it be removed, if that makes sense, it's just that asinine.

I hope this helps.

gemt:
I'm in no position to reject or accept any changes here. Only reason I ask is I'm sure those that will accept or reject the changes would be curious what it's based on, as am I. And while this is probably a more correct solution than the current one, it may still not be **the** correct solution, in which case, if the sources are shared, someone else might be able to further improve it more easily.

Blackbear123:
[This link]([](http://web.archive.org/web/20061115223930/http://forums.wow-europe.com/thread.html?topicId=14381707&sid=1)) from the old wow forums seem to backup what the original poster said about the 5,60 pct. hitchance and the missing 0,6. 
"Base miss chance is 5%. If you duel weild your white attacks (auto attacks) suffer a further 19%. All your +hit gear reduces this number. For every point the mob defense is higher than your weapon skill, miss goes up by 0.04. "

The post also states that "The calculation of glancing rate is (NPC lvl – YOUR lvl – 1) * 15 + 10"
Reduced damage from normal attacks as follows:
Normal damage - ((Defense skill – Weapon skill - 5) * 3%)

ssishtar:
To modify the current +skill/glancing system based on nothing, it's beyond idiotic. There is zero evidence to support this request, not even rumors.

femdead:
evidence please

Phoosy:
Echoing @gemt here - the actual evidence of this existing would be interesting to review beyond the "rule of thumb" on the percentage being picked up in other places, given there doesn't seem to be anything that indicates a presence of a contrary reason why it's this way.

blubba22:
"Hitcap with 304 weapon skill (Maladath, Anubisath Warhammer, etc.) is lowered from 7.40% to 7.00%, effectively 1%
Hitcap with 305 weapon skill (racials) is raised from 6.00% to 6.60%, effectively 1%
Hitcap with 307 weapon skill (Edgemasters) is raised from 5.80% to 6.40%, effectively 1%"

Can somebody explain to me why all these changes are "effectively 1%"?

ssishtar:
because the OP is an idiot.

sickheals:
@Phoosy This is because of the way the percent chance to miss works, it rounds down to the nearest whole number. So 5.80% chance to hit is 5% chance to hit, and 6.40% chance to hit is 6% chance to hit, so it crosses the whole number barrier and becomes essentially 1% chance to hit. Most of his argument hinges on this presupposition. I am in no place to say it is correct, though.

GuybrushGit:
@blubba22 there's no rounding going on 5.80% is 5.80%, not 5%.

dv3rgur:
The jump makes perfect sense, attacking a mob that has a level that is suddenly a different colour should be "jump" harder, i.e. going from weapon skill/defense difference of +15 to +10. I really hope this retarded argument wasn't the reason this dumb change was made.

PR #762, opened 2017-05-18
Fixed glancing blow calculation

Cephel:
The glancing blow calculation currently contains numerous "magic numbers" based on empirical research, which produce odd behavior in some edge cases.

This graph shows the currently working formula for the reduction in damage for a glancing blow (Zoom in for full detail):

The actual damage variation is very low and caps off at an unreasonably high value, as such almost no actual reduction takes place even when massive skill differences are taken into account. Additionally, no damage variation in between each glancing blow is present, when well documented and researched proof showed otherwise.

As such the corrected formula now introduces the damage this damage variation as shown below:

This formula is based on the 0.91/0.99 formula that has massive research and merit behind it and shows correct damage variations (thin bars) and average reductions (dot in the middle of each bar). You may note that this formula adds much harsher penalties for glancing blows at extreme weapon skill differences, this is intentional.

### DIFF
--- a/src/game/Objects/Unit.cpp
+++ b/src/game/Objects/Unit.cpp
@@ -1710,61 +1710,20 @@ void Unit::CalculateMeleeDamage(Unit *pVictim, uint32 damage, CalcDamageInfo *da
             // int32 victimDefenseSkill = pVictim->GetDefenseSkillValue(this);
             int SkillDiff = 0;
             SkillDiff = pVictim->GetDefenseSkillValue(this) - GetWeaponSkillValue(damageInfo->attackType, pVictim);
-            float reducePercent = 1.0f;
-            // (Youfie) Formule de calcul de la réduction de dégâts des érafles, influencée en pré-BC par le +skill au delà de [niveau du joueur * 5]
-            // Formule d'Athan retenue et supposée comme Blizz-like au regard des multiples sources et tests concordants
-            // float reducePercent = 1 - (5*(pow(2,(victimDefenseSkill/5) - (attackerWeaponSkill/5) - 1))/100);
-            // Après tentative d'implémentation "propre" de la formule et de nombreux échecs, mise en place de celle-ci après calcul manuel des différentes valeurs
-            // cf. http://nostalrius.org/forum/viewtopic.php?p=43964#p43964 pour infos et sources
-            if (SkillDiff >= 15)
-                reducePercent = 0.6500f;
-            if (SkillDiff <= 0)
-                reducePercent = 1;
-            switch (SkillDiff)
-            {
-                case 14 :
-                    reducePercent = 0.7018f;
-                    break;
-                case 13 :
-                    reducePercent = 0.7469f;
-                    break;
-                case 12 :
-                    reducePercent = 0.7860f;
-                    break;
-                case 11 :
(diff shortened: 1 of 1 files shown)


GuybrushGit:
>when well documented and researched proof showed otherwise

If this was actually well researched in vanilla it would have been changed ages ago.

What you implemented was confirmed by Landsoul during WotLK so we can assume it to be true.
Given the lack of data from previous expansions I think its safe to assume that this how glancing should work and its a much better system than the "magical numbers" we currently have.

Cephel:
For interest, this is how the two formulas look overlayed, as you can see they very closely match in the small skill difference variation range, but diverge at the extreme ends, especially towards the lower end:

dagochen:
So above 310 weaponskill it wont change anything....sounds strange....bosses have 315 defense....doesn't seem reasonable to make it capped at 310 then? the thing is nobody got more than 310 weaponskill that time so people said "there is no sense in getting more weaponskill!" and now everybody believes the cap is at 310 :D

Cephel:
Note that this is just glancing blows and note that this also heavily affects how fighting mobs while leveling.

Capping it at 315 practically dooms every single warrior that wants to fight a mob 3 levels higher than him, something that happens quite often while leveling and is a huge part of why the experience is so painful under the current formula.

The chance for a glancing blow also scales with the level difference (not skill), so +3 levels gives you your 40% glancing blow chance, but +2 levels only gives (pure guesswork) 20% chance or something.

gemt:
Interesting stuff, but you repeat "very well researched" a lot, but reference no sources. I know you have them, my point is simply, would it be possible to add some of them for completeness?

Cephel:
Most of this was researched (finding the research that is) on my part and one other person a few weeks ago, I can try to find our sources again, but it basically boils down to this:

- Finding sources for damage variation between each glancing blow (lower and upper limit) is trivial, dozens of people reported this being the case and it was as simple as walking to any mob a few levels higher and noting the damage spread of glancing blows. Since this is effect is currently missing from the code and improvent atleast in this regard must be made.
- A blue post confirming the exact AVERAGE damage reduction for a later expansion was found. The blue post confirms one specific reduction case, I think it was 70 vs 73 reduction on average. While this coefficient is confirmed, no information about the scaling exists in this bluepost, this is very likely where the source for the current magic numbers came from, plotting any random geometric progression that hits 100% on the top end and goes through the confirmed coefficient.
- To iterate on the last point, you may note that the new formula hits the exact -15 skill point as the coeffient does, if I remember correctly, that's because that's the one confirmed value and as such, any formula must fit this value in this specific case correctly, as it was blue post confirmed.

In reality, the changes aren't very drastic and I'd almost consider this patch a cosmetic change that only affects fringe cases (note the fitted plot between the two formulas).

BeerfishWoW:
This looks interesting, though it'd be interesting to actually see the plethora of resources that you've found to come up with these changes.

ssishtar:
This graph is bull and also it's poster is complete bull. I've been raiding at 315 and 310 in vanilla retail and I parsed tons of logs and everyone knew it was a clear difference of glancing blow valors between 310 and 315.  Sorry, but the glancing blows system works perfectly, as it should. Stop missinforming.

PR #764, opened 2017-05-18
PvP: Add config options to allow for accurate PvP item/zone requirements

StadenElysium:
- Entering rank restricted areas should be based on current rank, not highest achieved (lifetime) rank. 
- Equipping PvP items prior to patch 1.6 should be based on current rank.
- Purchasing new PvP items should be based on current rank.


strudelkuchen:
> Entering rank restricted areas should be based on current rank, not highest achieved (lifetime) rank.
This better not apply to Anathema before all the old rank 14 people get (end of this month) their items.

Hosq:
How the fuck does this get "fixed" in a matter of hours while there are ton of actual bugs to work on

StadenElysium:
Nothing's changing on Anathema.

src/game/Handlers/MiscHandler.cpp
```diff
@@ -805,14 +805,18 @@ void WorldSession::HandleAreaTriggerOpcode(WorldPacket & recv_data)
...
         return;
 
+    auto playerRank = sWorld.getConfig(CONFIG_BOOL_ACCURATE_PVP_ZONE_REQUIREMENTS) ?
+        GetPlayer()->GetHonorMgr().GetRank().visualRank
+        : GetPlayer()->GetHonorMgr().GetHighestRank().visualRank;
+
     if (!pl->isGameMaster())
     {
         // (Hack) : Entree dans les zones de recompenses JcJ
         if (Trigger_ID == 2527) // Hall des Champions
```
Inifield: This whole block can be moved to DB, Only need to implement a new condition "pvp rank requirement" and you're good to go

StadenElysium: I considered it but I didn't think it was much better.

dagochen:
>>Entering rank restricted areas should be based on current rank, not highest achieved (lifetime) rank.

I always thought it would be different. Maybe I am wrong, but I think I can remember people could still enter officers lounge after getting demoted. I already searched for sources about that issue, but never found anything. What sources you got so far?

strudelkuchen:
> Nothing's changing on Anathema and the 1.6 equip change isn't being used anywhere either.

It seems it was enabled on Anathema after all.
This is an issue for old R14 rankers. They can't enter rank restricted areas anymore to get their newly added weapon(s).

PR #768, opened 2017-05-18
Add core scaling support to aura 56 (SPELL_AURA_TRANSFORM)

### LINKED ISSUE #662: "Thekal's Grasp" and "Arlokk's Grasp" Set Bonus Buggy
The proc gives you a 300AP buff (working as indended) and transform you into a tiger. 

If the Setbonus Procs and you transform into a tiger **your size should increase a lot**!

As you can see here:

https://www.youtube.com/watch?v=K87jbP77P8Q

https://www.youtube.com/watch?v=aIy_wZCDo5o

**On Elysium you still have only the normal player size in tiger form**. very boring :(

(Tested on Darrowshire)

StadenElysium:
Fixes #662


WobLight:
after  #810, checking for scaling is no longer necessary and `apply == fasle` case doesn't need handling.

PR #775, opened 2017-05-19
Fixed gossips for all class trainers

brotalnia:
I went through every single class trainer and corrected their gossip text. All of them except the warlock trainers had "static" gossips assigned in `npc_gossip`, instead of a gossip menu, which meant they always said the same text without taking into account the class of the player. Also some of the rogue trainers had a core script to give the player an "Elegant Letter" which is for one of the rogue quests, but that was overwriting the gossip menu assigned to them in `creature_template`, so i removed that script from the core and implemented it in the database, like in mangoszero.


StadenElysium:
:+1:

PR #781, opened 2017-05-19
Update to MySQL 5.7.18

some2:
This updates the MySQL client libs (libmysql) from version 5.1.49 to 5.7.18 (current)


brotalnia:
Does that mean we won't be able to use any version of mysql older than 5.7 to host the database?

some2:
No. Probably this client library will even work with MySQL 4.X/3.X servers. So this will still work with your 5.X servers as long as you dont use features introduced in 5.7 in the rest of the code.

These commits have already been merged in the mangos fork cmangos.

PR #784, opened 2017-05-19
Update to OpenSSL 1.0.2k

some2:
Update from OpenSSL 1.0.0d to OpenSSL 1.0.2k


some2:
You should really consider to merge this. At least for security reasons.

PR #798, opened 2017-05-20
Improvements for ScriptedPetAI, Timbermaw Defender and Arcanite Dragonling

brotalnia:
ScriptedPetAI fixes:
- Scripted pets will now avoid breaking CC effects. If the target has a crowd control effect that breaks on damage, they will switch targets or stop attacking until the CC effect is removed. Does not apply to aggressive pets.
- Somebody was using tabs, so i replaced them with spaces.

Arcanite Dragonling:
- The spell timers were given a random value starting from 0, which led to casting the same spell several times in a row sometimes.
- Flame Buffet was being cast as **triggered**, so it was instant cast. The spell is not supposed to be instant.
- According to thottbot comments, the dragonling was defensive until 1.9. I added a check for that.

Timbermaw Defender:
- The healing touch used was scaling with level, so it was healing for about 2k. I replaced the spellid with 5188, which is the player version of Healing Touch rank 4, since i didn't find any suitable npc spells that didn't scale with level.
- If the owner doesn't need to be healed, it will look for other friendly targets that are below 50% hp.
- If the target is immune to nature spells, it will stop casting lightning bolt and melee it instead.


src/scripts/world/npcs_special.cpp
```diff
@@ -1433,7 +1438,7 @@ CreatureAI* GetAI_npc_arcanite_dragonling_dragonling(Creature* pCreature)
 ######*/
 enum
 {
-    SPELL_HEALING_TOUCH = 23381,
+    SPELL_HEALING_TOUCH = 5188,
```
Tobschinski: 26097

Jonathanfy:
well done, seems correct.

PR #871, opened 2017-05-26
don't clear combo points for cat forms on target change

WobLight:
Source:
https://youtu.be/uRQw9Y04elo?t=68


QAkvastskaft:
I'm quite sure this is invalid. Correct me if I'm wrong
Here is a video originally uploaded here: http://www.warcraftmovies.com/movieview.php?id=27552

At https://youtu.be/tXgsBdLQfvU?t=186 you see that combo points build on the Tauren stay when shifting forms.

Here https://youtu.be/tXgsBdLQfvU?t=941 you see 2 combo points being built on the Undead.
A few sec later he switches to the Orc and changes to bear and then belf (neither of which removes points) he then switches back to Undead and and no pints are there. 
https://youtu.be/tXgsBdLQfvU?t=964

--> Points gets removed as suggested by patch notes.

WobLight:
I've added the video just for reference, I've played feral from vanilla to cataclysm in retail.

Imbaslap:
incorrect source.
all rogues/druids do not drop combo points when fighting a player's pet/player.
the hunter's pet does not count as a combo point drop as it is still attached to the current targeted player.

also ignoring patch 1.4 fix http://wowwiki.wikia.com/wiki/Combo_point

Lorilay2:
wat

WobLight:
The video is indeed invalid, it may looks like the combo points are preserved, but it's just that the druid is using claw very quickly after changing target back to the hunter. #885 shows better evidence.

adura90:
Rogue combo points should not be lost when switching targets. Only if a new combo point is added to a different target are the combo points lost. You can see in this video from World of Roguecraft where the rogue fights the infamous Maydie that he does not lose his combo points when he un-targets him after blinding him. [ Time: 0:51 ] [ Video: https://www.youtube.com/watch?v=d7JYsuRD0nY ] [Note: This video is dated from 2007 and I'm certain that we all know the WoRC series. Maydie was, if I remember correctly an infamous Grand Marshal from the US server Arthas. Popular for his video with the Rammstein Feur Frei soundtrack. If this isn't enough proof, I have more videos lined up for you guys.]

Lorilay2:
Do you have one where someone targets a different person and then targets their original target with combo points intact?

WobLight:
Was that a replay to a deleted comment, was it?

Lorilay2:
Yes

PR #885, opened 2017-05-27
Druids should also drop combo points on target change

WobLight:
reverting #871 won't do, druids would not reset combo point when not in cat form.


brotalnia:
Wait what? So they are supposed to lose combo points on target change?

WobLight:
I've found a video showing that (I've lost the link, I'm searching for it again)

WobLight:
https://youtu.be/mTuymTgJzd8?t=631 combo points are near the level of the target
https://youtu.be/mTuymTgJzd8?t=907

WobLight:
Just to clarify, I though druids would reset combo point when adding a new combo point to a new target, not right on target change, but that seems to be true only after 2.0.3:  
*"Combo points added to a target should not be reset until you add a combo point to a different target. "*  
that was in rogues section, but looks like it was true for druids too.

Issue #870, opened 2017-05-26
Expose weakness (Hunt T2 8/8) consumes an arrow on proc


Zyrkmeister:
When the hunter t2 8/8 set bonus procs and expose weakness is applied to the target, the hunter loses an extra projectile. Pretty sure this isn't intended.

Video: https://www.youtube.com/watch?v=Z2fsum6UNGA

Edit: 
-Tested enequipping arrows after firing a shot but before the arrow landed. Expose weakness procced, but no arrow was consumed.
-If multishot procs the effect, every single debuff consumes an arrow. So firing one multishot with 3 expose weakness procs consumes 1+3 arrows. 
Edit2:
I implemented workaround fixes to YaHT for annoying autoshot timer mess ups. Whisper Zyrkmeister ingame if you're interested.

Lurkkihunter:
What's really annoying about the expose weakness proc is that since you lose an arrow, it resets autoshot timer in addons like rais and YaHT.

PR #945, opened 2017-05-31
Fix to Expose Weakness taking ammo

GuybrushGit:
https://github.com/elysium-project/server/issues/870 **Expose weakness (Hunt T2 8/8) consumes an arrow on proc**

Also did the same for Net-o-Matic. Found no other ranged spells that shouldn't take ammo.


thepparker:
Probably worth making them not proc any range skills either, similar to blind.

Issue #670, opened 2017-05-10
Badge of the Swarmguard proc rate


Raziya-Anathema:
Hello! I'm here to report a bug, I hope I'm in the right place.
https://elysium-project.org/bugtracker/issue/4413

The Badge of the Swarmguard apparently has an incredibly low proc rate on Anathema. As you can see from the videos from retail vanilla in this bugtracker thread, the proc rate is actually supposed to be much, much higher. In my own experience, and the experience of several other hunters who also have the trinket that I've spoken with, we almost never get six stacks even by the end of the use effect...

Here are the videos in case you have trouble accessing the bugtracker thread:
https://www.youtube.com/watch?v=6DQPytnxvlo
https://www.youtube.com/watch?v=8dFsuv9dta0
https://www.youtube.com/watch?v=WJWZ8jcswj0

And here is an archived thread from Thottbot where you can see a ton of comments about the high proc rate and functionality:
https://web.archive.org/web/20070510001706/http://thottbot.com/i21670

Here are a couple of notable comments:

"**Great for hunters
Score 1.2 Vote: [-] [+] by dabba, 1.1 years ago**
A guildie of mine got it, it proc almost 100% on each attack, If he multishot and hit 3 target he get 2-3 proc. Awsome whit this trinket + rapid fire after 10 sec, then FD when it is on Cd and change trinket. "

"**Proc rate and effectiveness
Score 4.4 Vote: [-] [+] by Sama[of]Warsong, 1.0 year ago**
Not sure what the proc was like before but I can confirm that as of the current patch 1.10:

- It does stack with Sunder
- The proc actually gives you a buff which can carry over from one target to another
- Considering an 80% proc rate, it stacks up almost unbelievably fast"

I hope that's enough to go off of. Thank you for looking into this!!

GuybrushGit:
Research by Bazzil from Kronos:

https://www.youtube.com/watch?v=WJWZ8jcswj0
10 attacks = 6 stacks = 60% proc-chance

https://www.youtube.com/watch?v=6DQPytnxvlo
18 attacks = 6 stacks = 33% proc-chance

https://www.youtube.com/watch?v=8dFsuv9dta0
7 attacks = 5 stacks = 71% proc-chance

https://www.youtube.com/watch?v=5MAZkK_xdFg&feature=youtu.be&t=1m43s
11 attacks = 6 stacks = 54% proc-chance

https://www.youtube.com/watch?v=sN1K21Nth1s
[...]

PR #947, opened 2017-05-31
Badge of the Swarmguard proposed proc rate

GuybrushGit:
https://github.com/elysium-project/server/issues/670 **Badge of the Swarmguard proc rate**

Set it at 15 PPM, which is 75% for slow weapons (3.0) and 50% for fast weapons (2.0).
It seems about right from looking at the videos and comments but there isn't any research from retail to confirm this.


dabadoo:
Has this fix been applied to the servers yet? The proc rate is still horrible

PR #967, opened 2017-06-03
OpenSSL 1.1 support

ruscur:
These changes allow building on systems with OpenSSL 1.1 instead of 1.0.  This isn't too important right now (only bleeding edge Linux distros ship with it), but as time goes on more and more build environments will ship with OpenSSL 1.1.  You can't just point to OpenSSL 1.0 either because libmysqlclient links to OpenSSL 1.1 for me.

Tested on Arch Linux (OpenSSL 1.1) and Ubuntu 16.04 (OpenSSL 1.0).


StadenElysium:
This breaks Windows (VS) support.

ruscur:
Guessing in the CMakeLists for shared (or something) there needs to be a specific include for openssl on windows

PR #969, opened 2017-06-03
Group members visibility after entering dungeon together v2

sipertruk:
v1 was overdoing.
v2 is inspired from how trinitycore does it.


Kittnz:
Then credits where credits are due.

sipertruk:
Correct, If it works on live I'll add a comment.

Kittnz:
@sipertruk thank you

PR #971, opened 2017-06-03
Added script for pet Cannonball Runner

brotalnia:
The pet Cannonball Runner had no script, so it was auto attacking and chasing enemies. I scripted it and did some other trivial changes in the code for the Timbermaw Ancestor and Arcanite Dragonling that were bothering me.

UNIT_FLAG_DISABLE_ROTATE wasn't implemented server-side, so creatures with it were actually changing orientation. It seems like this flag was completely unused. I added a check for it in SetInFront as the Cannonball Runner is not supposed to rotate and only fires at enemies in front of it.

Also added an additional optional parameter in SelectRandomUnfriendlyTarget to let it return only targets in front of the unit.


src/game/Objects/Unit.cpp
```diff
@@ -9723,7 +9724,7 @@ Unit* Unit::SelectRandomUnfriendlyTarget(Unit* except /*= NULL*/, float radius /
     // remove not LoS targets
     for (std::list<Unit *>::iterator tIter = targets.begin(); tIter != targets.end();)
     {
-        if (!IsWithinLOSInMap(*tIter))
+        if ((!IsWithinLOSInMap(*tIter)) || (inFront && !this->HasInArc(M_PI_F / 2, *tIter)))
```
thepparker: Note that checking M_PI_F/2 only checks a 90 degree cone in front of the unit, not a full 180 degrees. PI = 180, 2PI = 360.

brotalnia: Well that seemed to work best for this case. Cannon balls don't fly sideways after all. If this is needed somewhere else, then the angle can be added as a parameter as well.

thepparker: This is true, just thought I'd point it out in case you implemented it with the expectation of searching 180.

PR #978, opened 2017-06-04
Several improvements for Darkshore quests

brotalnia:
Even more fixes for Terenthis' quests. 

- **Sentinel Selarin script**
There was no need to search for the sentinel by id in the waypoint script to set her flag, because it's executed with her as the source.
- **Quest: A Lost Master 1**
Terenthis will now clap when turning in the quest.
Removed the wrong quest end script and replaced it with the correct one.
- **Quest: A Lost Master 2**
Added missing quest start script. There is one emote text.
- **Quest: Escape Through Force**
Terenthis should bow when turning in the quest.
Added missing RP event when reaching the end of the escort. Grimclaw should turn around when he hears Volcor and run to him, then they each have 2 lines of dialogue before despawning.
- **Quest: Escape Through Stealth**
Terenthis should bow when turning in the quest.
- **Quest: Fruit of the Sea**
Gubber Blump should cheer when turning in this quest.
- **Volcor**
Added missing gossip menu.


m-laniakea:
Love your attention to detail, thank you.

PR #987, opened 2017-06-05
esES locale file update

Locario:
File to update esES locales loc6 rows in tables: creature_ai_texts, db_script_string, locales_creature, locales_gameobject, locales_gossip_menu_option, locales_item, locales_npc_text, locales_page_text, locales_points_of_interest, locales_quest and mangos_string.


Locario:
btw only some tables are of MaNGOS r2 eswow2.org

They have all the translations, only need to match the correct ID's, they will be another pull request.

PR #993, opened 2017-06-05
Implemented DetailsEmote handling for quests

brotalnia:
I noticed that the DetailsEmote on quests wasn't being displayed at all. That's the emote when the quest is being offered to you. So i looked into the code, and it turns out that it wasn't being sent at all. Apparently it's either not part of the SMSG_QUESTGIVER_QUEST_DETAILS packet, or the packet structure is not the same as in wotlk, and nobody bothered to figure out where the emote fits in vanilla. I came up with a simple workaround to send the emote to the player using a delayed event.

See the gif if you don't get what I'm talking about.
ezgif-1-34a527cd98

### DIFF
--- a/src/game/GossipDef.cpp
+++ b/src/game/GossipDef.cpp
@@ -536,6 +536,14 @@ void PlayerMenu::SendQuestGiverQuestDetails(Quest const *pQuest, ObjectGuid npcG
     */
     GetMenuSession()->SendPacket(&data);
 
+    // The DetailsEmote is not part of this packet in vanilla.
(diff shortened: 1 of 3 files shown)


Kittnz:
FYI i don't think this was the case in vanilla, https://youtu.be/oIeRYp956q4?t=4

brotalnia:
It did exist in vanilla for sure. Not every quest has a DetailsEmote though, that's normal.

Look at Joana's leveling videos. Even the first quest has one.
https://youtu.be/jTSyexUNN9E?t=1m21s

Issue #977, opened 2017-06-04
Hunter's mark displaying on minimap for non-party members. 


ChristianTucker:
I haven't tested to see if this occurred across faction, (ie: being able to see if a Horde hunter marks an Alliance player, and being able to see an alliance ally marked by an enemy hunter) however I've noticed several times that marked targets appear on the minimap when they are marked by a hunter of the same faction. 

I only played during the final days of Vanilla before TBC was released, however I don't recall this ever being the case. Earlier today in the starting zone my map was filled with dots from hunters marking different targets while questing. I'm fairly sure it should only show the mark if you're in a party with them.  (Or actually if you're the hunter, and it shouldn't matter otherwise)

WobLight:
https://youtu.be/4Pp9Kp1F3KY?t=24

Lorilay2:
This was always the case since the dawn of time.

WobLight:
Why did you closed it? This is valid and should be closed by Elysium.
Other references:
https://youtu.be/4Pp9Kp1F3KY?t=777
https://youtu.be/HxfN9f88P0I?t=5

Actually, looks like the mark is visible only to the hunter itself
https://youtu.be/HxfN9f88P0I?t=28

notpentadactyl:
> This was always the case since the dawn of time.
This is how I remember it as well. Might have been changed in a patch.

WobLight:
I can confirm that team could see marks in TBC, they changed it for arenas imo,

WobLight:
For anyone still skeptical
https://youtu.be/2bIvcPRkoUI?t=10
https://youtu.be/2bIvcPRkoUI?t=233

ChristianTucker:
@WobLight  I have re-opened it. I was testing this on other servers from Vanilla -> WoTLK as it was stated that this wasn't an error. After some testing, I can confirm that the mark is only supposed to be visible to the hunter.

taurenrogue:
I'm convinced that there should be no map icon for anyone but the hunter, but there's still no proof that marked target should be invisible to the party...

WobLight:
Why? Proof linked just 2 comments above

WobLight:
oh wait, you mean invisible when stealthed?

taurenrogue:
> oh wait, you mean invisible when stealthed?
Yes of ofc invisible via invis pot/succubus invis / cloaking device :)

WobLight:
[...]

PR #1002, opened 2017-06-06
Limit unit tracking visibility to the caster

WobLight:
This fixes https://github.com/elysium-project/server/issues/977
It's so bad it's almost an hack, but there's no other way afaict.

### DIFF
--- a/src/game/Group/Group.cpp
+++ b/src/game/Group/Group.cpp
@@ -313,7 +313,7 @@ bool Group::AddMember(ObjectGuid guid, const char* name)
         return false;
 
     SendUpdate();
-
+    updateSubGroupTracking(GetMemberGroup(guid));
     if (Player *player = sObjectMgr.GetPlayer(guid))
     {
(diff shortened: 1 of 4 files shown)


thepparker:
I'm pretty sure the hunter's party should see the tracking from hunter's mark

WobLight:
Unfortunately it's very hard to find videos about that, the only one I've found it's this one:
https://youtu.be/HxfN9f88P0I?t=28
You can see there's a hunter in the player's party and the target is marked, yet nothing is shown on the minimap. Of course it could be a mark of another hunter, but I can't find a better video. There was one with world PvP and and hunter in party, but the marked one was near the player so the minimap dot might have been overlapped.

WobLight:
collected one more
https://youtu.be/2bIvcPRkoUI?t=10
https://youtu.be/2bIvcPRkoUI?t=233

thepparker:
Ok, well 3:48 confirms it then. Only visible for the hunter (or hunters?)

PR #1004, opened 2017-06-07
Added gossip for some NPCs.

webdkpthroaway:
Added gossip for Andorgos and Kandrostrasz in AQ40.

You can view their text in the current live version of WoW, as it has remained unchanged. 
https://youtu.be/tGv1UxcBX7s?t=69


webdkpthroaway:
Modfied in accordance with comment above.
Left the Ouro teleport text out, afaik it was added in one of the later patches.

PR #1032, opened 2017-06-09
Map::Update: prevent small lag spikes from stacking delay on timers.

WobLight:
This adds a buffer to allow Map::Update to recover time lost in `diff > maxDiff` check.
This can be potentially harmful for timers' duration to suddenly shorten,
to prevent that, a maximum buffer capacity is to be set (0 to disable the buffer).

Currently setted to 0 by default, ~I'd suggest to test a value around 300/500ms, but an even smaller value could suffice.~ Due to some "issues" with map updates, the buffer could be consumed much faster than intended, so I'd actually suggest much smaller values (50ms for instance).
```
   ↓map update start           ↓map update start
[  |....          ][           |.... ]
↑_________________↑↑_________________↑
 world update       world update
   ↑___________________________↑
    map update diff
↑__________________↑
 world maxDiff
```


dagochen:
What would be the benefit for players? What can they actually see as improvement?

WobLight:
It should increase some timer's precision, preventing them from slowing down. My observation are based on BG timers such as time to capture a base and korrak spawn time.

dagochen:
Does it affect sd scripts timer as well?

WobLight:
"sd" stands for?

brotalnia:
I assume ScriptDev.

WobLight:
It depends on the way the timer is implemented, but most likely it does.

Issue #1033, opened 2017-06-09
Revantusk (Hinterlands) issues


Plus5Mana:
Noticed a few  possible bugs with Revantusk:

Huntsman Markhor - shows as a stablemaster, but no stabling is available when you click on him (quests on him do function normally though)

Otho Moji'ko - Cooking supplier, shows a yellow quest marker over his head, but no quest is available to be selected from his menu

Inn - rested area is only on second floor of the inn, and small parts of the platform away from the innkeeper (lard)

DonkeyKongRacer:
Huntsman Markhor have always had "Stable Master" as a subname according to most DBs on google. If you can find a reliable source showing that it shouldn't, I'd gladly fix it (although of course it's strange, but if it's blizzlike ...). but I'm pretty sure it has always been like this.

Otho Moji'ko isn't supposed to give any quest, it seems his npcflag is messed up. I'll fix this.

Inn - currently have no idea if this is really a issue, from what I know (I might be wrong) the inn areas are loaded from the DBC files. I know some DBC files have been wrong in the past (at least on zeth'kur side) and I have no idea why (raised this issue on  the dev discord back then) , but I don't see it being an error on elysium's part this time.  Since DBC files are extracted from the client, I'm pretty sure this is blizzlike.

Plus5Mana:
Regarding Markhor, It seems he's missing his feature set. If I go to any other stable master, i can click on them and use their stabling features (as a hunter) - for instance, Hekkru in Stonard. But with Markhor, I get no options to either stable/retrieve a pet - only to participate in the quests he has.

DonkeyKongRacer:
Yeah sorry I misunderstood you. I'm fixing both of these NPC.
As for the rest trigger zone, it's indeed very weird, but afaik it's not a bug considering how it's handled by the core ? It would rather be a client issue. Could have been like that in retail.

thepparker:
Resolved

PR #1038, opened 2017-06-09
Fix wrong NPC flags for two Revantusk NPCs

DonkeyKongRacer:
Fixes an issue with NPCs listed in #1033
This doesn't fix the problem with buggy Inn rest zone stated in that issue. Afaik the trigger zone for rest zones is loaded from the DBC files, and since it works pretty well anywhere else, it might have been like that in retail ? So not really an issue in my opinion.


thepparker:
You need to use different branches for each separate issue you want to address. This PR includes changes from another issue that are in your base development branch.

PR #1040, opened 2017-06-09
Many quest start & end script fixes

brotalnia:
I noticed some quest_end_scripts were added more than once, so npcs where saying the same thing multiple times, too many mobs were being spawned, etc. But as i went to fix the duplicates, i noticed a lot of flaws in the scripts, like missing texts and emotes, missing waypoints, wrong timing and so on, so i went ahead and fixed them to be more blizzlike. I also added the correct localized texts for all the quest scripts i fixed.

- **Fields of Grief 2**
Everything was added twice, some of the timing needed adjustment, and the waypoints were missing.
- **Quest: A New Plague 2**
Script was added twice, and Captured Mountaineer didn't say his line. Also he didn't have a mug in his hand when drinking the poisoned drink you give him, so it looked like he was drinking an imaginary drink. He is supposed to have a mug in his hand when he drinks, which can be seen in old retail videos.
- **Rite of Vision 2**
Both texts were missing, and the line to make him cast his spell was added twice.
- **Altered Beings**
His script consisted of one emote after 15 seconds (added twice ofc), so basically the quest didn't have a script. I added the proper script.
- **Samophlange**
Everything was duplicated. The dialogue was missing, timing needed adjustment, and the spell he casts didn't have the Samophlange gameobject defined as the target in the database so it produced an error in the server console whenever he cast it.
- **The Absent Minded Prospector 4**
Aman was spawned in the wrong spot, and Archaeologist Flagongut was moving to some random place. There was no dialogue at all. So basically the whole thing was wrong.
- **The Tower of Althalaxx 4**
Everything was duplicated, so it was a cacophony of text. The timing needed adjustment, Delgren the Purifier wasn't actually facing the orb when doing his smashing animation, and the spawned npc Sargath had his text added as an EventAI script, instead of handling that in the quest end script. Also all of Delgren's dialogue was missing.
- **Nugget Slugs**
The script was duplicated, Wenikee Boltbucket wasn't actually facing the Samophlange when hitting it, and her dialogue was missing.
- **Proving Allegiance**
This one wasn't duplicated, but the text was missing.
- **Report to Orgnil**
This quest had no quest end script at all.
- **The Prodigal Lich Returns**
Gunther's Visage didn't walk to Bethor. The npcs didn't face each other when talking. Gunther wasn't facing the player when speaking to him. Bethor was saying one of Gunther's lines. Timing was wrong.

I just want to remind everyone that most of the database tables having to do with scripts have no key, so the same line can be added multiple times. This is so that multiple actions can be performed at the same time in the script, but it leads to duplicates when people are not careful. You cannot use REPLACE to change entries in these tables, because there is no conflict when adding them more than once, so it's functioning as an INSERT statement. It just adds your entry a second time.


brotalnia:
Fixed end script The Prodigal Lich Returns as well. Made it identical to this retail video -> https://youtu.be/Bc1AyT9ATY8?t=7m44s

Issue #1052, opened 2017-06-11
Grave Moss spawn timers 1-2h+


Grimjoy123:
I've seen this issue being posted here before and just want to magnify it. The respawn timers should be 6-8min or something which is blizzlike.

 I currently try farming them on Zeth'Kur as the lone Grave Moss farmer, no competition, been camping both Ironbeards tomb and raven hill cementary and the fastest respawn I came across yet is 1h 30min.

There's literally NO grave moss ever on AH and people are even making videos on how to farm it in SM GY because there's 1 spawn in there thats guaranteed and u can do it 4 times in a hour. 

Now for things to go THAT far, it's clearly become an major issue especially with Naxxramas and Greater Shadow Protection Potion consuming soon coming up on Anathema.

Blizzlike: http://www.wowhead.com/item=3369/grave-moss#comments
Issue: https://elysium-project.org/bugtracker/issue/2551

stesve:
There is only vanilla comment on that page and it actually recommends farming them in SM.

> Farming for these for Loatheb has made me want to kill myself.

> But, if you are looking to farm them, I've found the best place to be in Scarlet Monastery. I'm getting about 3 grave moss and some kingsblood and fadeleaf every time i reset the graveyard.

Grimjoy123:
Alright, what about the missing spawns?

GuybrushGit:
Respawn is 30 mins but 2/3 of the nodes on the database are disabled, no clue why.

Grimjoy123:
LEGEND, thank you sir for helping out!

PR #1053, opened 2017-06-11
DB/gameobject: Enable Gravemoss Spawns

m-laniakea:
Around 57/86 Gravemoss spawns were disabled in the DB
https://github.com/elysium-project/server/issues/1052


m-laniakea:
Actually, these objects need to be pooled

Issue #1023, opened 2017-06-08
Shield Slam not triggering Chance on Hit


Yoshgit:
Extensive in game testing has shown that Shield Slam does not trigger main hand weapon procs.

http://eu.battle.net/wow/en/game/patch-notes/2-0-1
"Warrior "Shield Slam" and "Shield Bash" will no longer trigger a proc from an equipped weapon."

This implies that before 2-0-1, shield slam and shield bash were both triggering procs.

hckYosh:
Noticed that similar fixes have been proposed but denied. Now that there is indisputable evidence from Blizzard's own patch notes, a similar fix should resolve this problem.

[https://github.com/elysium-project/server/pull/187](url)

vindruvor:
"Warrior "Shield Slam" and "Shield Bash" will no longer trigger a proc from an equipped weapon." is below the section 'BUGS' so I dunno what u do in that case :o

hckYosh:
I understand that it is labeled under bugs, but the patch notes reflect that this has been a mechanic that existed from Shield Slam's inception until TBC. At the end of the day its up to the devs, but intentional or not Shield Slam behaved this way during its entire course in Vanilla wow.

vindruvor:
I think this topic is hard; to decide wether a mechanic that got defined as a bug in a fix after vanilla.

Like for example "Fixed an issue where the Warlock spell "Death Coil" gained double the benefit from the talent "Shadow Mastery"."

Should that mean that for anthema we should have 2x benefit from talent "Shadow Mastery" for death coil?

hckYosh:
I understand the point you are making. My point was that either way, there is a bug here. If shield bash was fixed to trigger main hand procs, there should be some consistency by applying the same fix to shield slam. If this was a mistake, then the shield bash change should be reverted. Like you say, past mechanics which were reclassified as bugs later is a difficult issue. As a fury warrior, these changes aren't even relevant to me and I stumbled upon this issue by mistake. Consistency is the key word here.

PR #1057, opened 2017-06-13
Shield Bash should not trigger main hand

Yoshgit:
Addressing #1023

Proof:
http://eu.battle.net/wow/en/game/patch-notes/2-0-1
"BUGS: Warrior "Shield Slam" and "Shield Bash" will no longer trigger a proc from an equipped weapon."


robmonte:
Why was this merged? The patch notes linked clearly show patch 2.0.1? Am I missing something?

Yoshgit:
It was labeled as a bug. It was never intended to proc main hand weapons.

robmonte:
There are many other bugged mechanics that are fixed in later patches that were left in the server on purpose because that's how they actually were in vanilla. Wasn't that the goal, to me as true to the real blizzard vanilla experience as possible?

Yoshgit:
I'm not related to the QA team in any way, but think the QA team follows the philosophy of not reproducing bugs, whether they happened in vanilla or not.

robmonte:
Interesting. I wonder when the stance changed on this? Because I'm pretty sure I remember them being in favor of leaving in vanilla bugs.

PR #1094, opened 2017-06-17
Fail Skinning attempts if the player isn't within loot range

ruscur:
On large corpses with disjointed models (i.e. Devilsaurs with their necks and tail sticking out), the client can think it's in range to skin without also being within the INTERACTION_DISTANCE required to
loot.  If this occurs, the SendLoot() call fails, no loot is retreived, and the corpse is no longer skinnable.

Make sure the player is within looting range of the corpse they're skinning, and if not, report that the spell was out of range.


thepparker:
Probably worth checking this in Spell::CheckCast rather than failing after the cast has completed

ruscur:
That does sound better, I'll try that.

PR #1095, opened 2017-06-17
Various quest related fixes & some ScriptMgr improvements

brotalnia:
A few quest related corrections and a small update to ScriptMgr.

- **Quest: Deaths in the Family**
This quest should not have a previous quest requirement.

- **Quests: Hallowed Scroll, Glyphic Scroll, Encrypted Scroll**
These quests have a wrong previous quest requirement. They should require The Mindless Ones, not a quest all the way in Brill.

- **Quest: Your Place In The World**
It has wrong RequestItemsText from another quest. It can't be seen since the quest has no item requirements, but it's still wrong.

- **Quest: Minshina's Skull**
Improvements for the `quest_end_script`. Corrected the timing and despawn delay, made it so the npc can't be spawned a second time while the npc is still there, made it spawned facing the player and added localized text.

- **Quest: The Prodigal Lich Returns**
I forgot to remove the quest flag from the npc when starting the event.

- **ScriptMgr: SCRIPT_COMMAND_TEMP_SUMMON_CREATURE**
Added the option to spawn the npc facing the source or target, and to set walk/run mode on spawn as is done in mangoszero.

- **ScriptMgr: SCRIPT_COMMAND_TERMINATE_SCRIPT & SCRIPT_COMMAND_TERMINATE_COND**
Use continue instead of return to avoid halting execution of scripts until next update.

- **ScriptMgr: SCRIPT_COMMAND_TURN_TO**
Refactored this one to reduce the amount of code and fix a bug where a certain combination wasn't working.


brotalnia:
- **NPC: Nazgrel**
He was missing his weapon.

- **NPCs: Kor'kron Elite and Deathguard Elite**
These should not be elite actually.

- **Quest: Find the Gems**
There was a typo in the text.

m-laniakea:
Deaths in the family [does](http://web.archive.org/web/20060325200439/http://wow.allakhazam.com:80/db/quest.html?wquest=354) have a prerequisite.

brotalnia:
That seems to have been a mistake by Allakhazam, as there is multiple evidence to the contrary.

Videos showing both quests being offered simultaneously:
https://youtu.be/DhTYPNsiKk4?t=6m50s
https://youtu.be/CVeijybDMN0?t=8m40s
https://youtu.be/8dNMBWh2wxY?t=9m27s

Questing guide from 2006:
http://www.wow-pro.com/node/656
> After you´ve arrived in Brill, get the two quests from Coleman Farthing next to you, Deaths in the Family and The Haunted Mills.

Another database:
http://wow.gamepressure.com/quest.asp?ID=854 <- No mention of a requirement.
But it should list previous quests when there is one:
http://wow.gamepressure.com/quest.asp?ID=846 <- Previous quest is listed for "Doom Weed".

m-laniakea:
Very well, thank you.

PR #1123, opened 2017-06-21
Starshards WoW 1.10.0

WobLight:
# [Patch notes](http://wowwiki.wikia.com/wiki/Patch_1.10.0)
*Starshards (Night Elf) - This spell now works like Curse of Agony, with lower initial damage which increases as the duration continues. Mana cost dropped approximately 30% on all ranks. Ranks 1 and 2 increased to 6 second duration. Damage increased approximately 17%.*

Coefficients taken from this [video](https://youtu.be/eBp_P_Pus7A?t=59) from 2007.

### DIFF
--- a/src/game/Spells/SpellAuras.cpp
+++ b/src/game/Spells/SpellAuras.cpp
@@ -1465,7 +1465,7 @@ void Aura::HandleAuraDummy(bool apply, bool Real)
                                 if (!player->InBattleGround())
                                     player->RemoveAurasDueToSpell(2584);
                         }
-                        return;                            
+                        return;
                     }
                     case 10255:                             // Stoned
                     {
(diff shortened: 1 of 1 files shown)


WobLight:
Needs adjustments, more math proved that coefficients are 0.667/0.667/1/1/1.333/1.333 of base tick damage (coeff * 1/6 of total) and spell power is added flat to all ticks (1/6 of total bonus per tick).

WobLight:
Curse of Agony and Starshards tick coefficients should only affect the base damage, not the total outcome, as observed here [starshards.xlsx](https://github.com/elysium-project/server/files/1096184/starshards.xlsx)

Kittnz:
How about adding a WOW_PATCH_110 check in there?

WobLight:
I'm generally against mechanism progression, it doesn't match client and brings confusion.
Starshards is a very good example, currently it works halfway, total damage and duration are from 1.10 already, but tick scaling wasn't implemented. If you override everything to have a pre 1.10, you will have Ranks 1 and 2 with wrong description in the client (in 1.10 duration was increased of 1 sec and damage about 17%).

Kittnz:
Fair enough, spells and such are 1.12 anywya

PR #1124, opened 2017-06-21
Transport fixes

### LINKED ISSUE #214: Disconnect on ship to Feathermoon: CONFIRMED
BUGGED BEHAVIOR
Disconnected when run in into Feathermoon ship.

 CORRECT BEHAVIOR
No disconnect.

 SOURCES
S1

Original post on bug tracker here: https://elysium-project.org/bugtracker/issue/1316

### LINKED ISSUE #437: Boat issues after the patch 2017.04.19.
Bat pet (Bloodseeker Bat) goes up tho hte air, to the top of the boat, teleports back to the owner on boat leave.
http://imgur.com/a/OZIow

And not every player gets dismounted on entering boat.

sipertruk:
- Workaround a bug in the client, sometimes when boarding it will keep sending CMSG_MOVE_TIME_SKIPPED like it's lagging. It happens more often with latency, low video settings, you could predict something will go wrong when seeing water through the boat's doors. Should fix #214
Probably related: https://elysium-project.org/bugtracker/issue/1669 https://elysium-project.org/bugtracker/issue/1316 

- By removing first and last node from the path generation, the events timing at the client and those calculated by the server become very close. This improves passengers visibility and maybe other things, you can know more precisely where it should be. These 2 nodes are not real waypoints ?

- After a while the client will destroy a transport on its own, so unlike it's refreshed on the way (by porting in the map) you could wait for it forever, Feathermoon-Savage Coast and Auberdine-Teldrassil were affected.

- Fix #437  better mount check when boarding + flying pet in boat is weird because they should all be grounded anytime when tamed.


sipertruk:
I think I'll add a Transport::SendUpdateTo for the extra update instead of using TeleportTransport at the same location.

PR #1130, opened 2017-06-22
Fix flurry ranks 1 to 4 proc

### LINKED ISSUE #327: Flurry activates on ranged critical hit: CONFIRMED
BUGGED BEHAVIOR
When you land a critical hit with a ranged weapon flurry activates. The talent clearly states "melee critical hit".

 CORRECT BEHAVIOR
Flurry should only activate on melee critical hits.

 SOURCES
Me?

Original post on bug tracker here: https://elysium-project.org/bugtracker/issue/2161

sipertruk:
Fix #327
It was fixed in patch 2.1.0.
https://www.engadget.com/2007/05/22/2-1-patch-notes-part-1/
[Flurry]: Ranks 1-4 will no longer trigger from ranged critical strikes.


thepparker:
It's not really a Vanilla mechanic if it was changed in 2.1, is it?

sipertruk:
The flurry tooltip for all ranks is only about melee crits :
"Increases your attack speed by XX% for your next 3 swings after dealing a melee critical strike."
Yet only rank 5 proc only on melee, this could be considered a bug.

sipertruk:
If you don't want it then close #327

PR #1155, opened 2017-06-25
Fix CoA and Starshards bonuses

WobLight:
Starshards and Curse of Agony change their base damage over time, but spell power bonus is the same for every tick.
To achieve that, we can't store total damage at cast time, but we have to store base damage and bonuses separately for later calculation.

### Coefficients
[starshards.xlsx](https://github.com/elysium-project/server/files/1100416/starshards.xlsx)
- Starshards samples are from early TBC videos
- CoA samples are from Vanilla videos

### Implementation Details
instead of the total damage,
Modifier now stores:

 Member     |  purpose
------------|-----------------------------------------------------------------------
m_base      | base points (e.g. damage) of the aura
m_bonus     | flat addition (e.g. spell power)
m_bonus_pct | percentage bonuses (e.g. talents +x% bonus damage)
m_used      | currently for shields assorbed damage (e.g. Power Word: Shield)
total(c)    | return value calculating bonuses (m_base * c + m_bonus) * m_bonus_pct
remaning()  | return total() - m_used, currently used for remaning shields value

pet_aura and character_aura table have been updated to store such data as well.

### DIFF
--- a/sql/migrations/20170625144515_world.sql
+++ b/sql/migrations/20170625144515_world.sql
@@ -1,0 +1,23 @@
+INSERT INTO `migrations` VALUES ('20170625144515');
+ALTER TABLE character_aura
+ADD COLUMN bonus0 INT AFTER basepoints2,
+ADD COLUMN bonus1 INT AFTER bonus0,
+ADD COLUMN bonus2 INT AFTER bonus1,
+ADD COLUMN bonus_pct0 FLOAT AFTER bonus2,
+ADD COLUMN bonus_pct1 FLOAT AFTER bonus_pct0,
+ADD COLUMN bonus_pct2 FLOAT AFTER bonus_pct1,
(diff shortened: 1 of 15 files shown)


WobLight:
Squeezed struct to minimal, description updated above.

WobLight:
now spell shouldn't loose value to rounding until reduction/absorb calculation. Values are dithered just before that.

ItsmeAga:
do you still have the links for those videos? Also it is a bit hacky, but instead of the rework one could simply set the spellpowerscaling in the DB to zero and manually add the spellpower bonus dmg in SpellAuras.cpp

WobLight:
I'll look for the links, anyway to calculate spellpower at tick time, you would need to save much more data (spell power and all modifier at cast time).

brotalnia:
There are merge conflicts in 2 files.

PR #1165, opened 2017-06-26
Script improvements for Stitches

brotalnia:
- Moved texts to database instead of hardcoding them in the script, and added localizations.

- Added some missing texts to the Town Crier.

- Made the Town Crier yell to the whole zone.

- Tweaked timing. Some of the watchers were spawning too early.

- Spawned Watchers should despawn when their corpse despawns, not immediately after being killed.

- Added missing RP event between Mayor Ello and Watcher Cutford.

- Tweaked Watcher Selarin's waypoints to try to avoid clipping through the fences along the road. Also she should stop at a point and wait for Stitches there, not run all the way to the end of the road.

- Added missing text when beginning quests Wait for Sirra to Finish and Translation to Ello.


Kittnz:
Looks good!

PR #1171, opened 2017-06-27
Add reward quest status

IlseElysium:
Instead of showing COMPLETE on the quest status for both when you have the requirements of a quest and when you have actually completed the quest, make it so that it shows REWARDED to the completed quests.


StadenElysium:
This is going to cause breakage in a few places and I don't think it's best solution to your problem.

The `QuestStatus` enum is listing states that the client cares about, whereas whether a quest has been rewarded is something that only the server cares about. It's not a big deal but it'd be nice to keep a distinction between them.

IlseElysium:
Here's your solution's way.

some2:
The Travis CI error already occurs before compiling your code.

Its a problem with the debian.neo4j.org repo (probably not the repo it self but with apt)

PR #1173, opened 2017-06-28
Set required level for quest "In Search of Thaelrid" to 18 (#1162)

### LINKED ISSUE #1162: Quest: In Search of Thaelrid level Requirement too high
Quest level is wrong. It is currently lvl 20+ and should be 18+.
http://www.wowhead.com/quest=1198/in-search-of-thaelrid
https://vanilla-twinhead.twinstar.cz/?issue=6508
If you look at the next quest in the chain, Blackfathom Villainy, it is level 18+. Horde can pick up Blackfathom Villainy at lvl 18 and obtain the wand. This is important for twinks. 
quest id: 1198

some2:
See https://elysium-project.org/bugtracker/issue/2491 (from 20 down to 18)

Fixes #1162

### DIFF
--- a/sql/migrations/20170628103600_world.sql
+++ b/sql/migrations/20170628103600_world.sql
@@ -1,0 +1,4 @@
+INSERT INTO `migrations` VALUES ('20170628103600'); 
+
+-- Set required level for quest "In Search of Thaelrid" to 18
+UPDATE quest_template SET MinLevel = 18 WHERE entry = 1198;


cocobrown23:
Can confirm this. This bug has been tracked for over a year now (old bug tracker) and still has not be solved. Huge imbalance for horde casters to have access to the wand at 19 while alliance does not.

PR #1238, opened 2017-06-28
Small fixes for Stormwind children

sipertruk:
Some were attack-able.

Donna and William drama, they were eventually overlapping due to how the grid updates and they are supposed to run. 
Setting one of them or both as active creature won't work because in that case the distance between them raise. So I did a creature_group.

#384 is needed for Donna's text to work because Map::ScriptsProcess always sets a target (William) for the chat.


brotalnia:
You are using `creature_movement_scripts` ids that are already taken. Free values start from 21 i believe.

sql/migrations/20170629002825_world.sql
```diff
@@ -0,0 +1,140 @@
...
+DELETE FROM `creature_movement` WHERE `id`='79721' and`point`='79';
+DELETE FROM `creature_movement` WHERE `id`='79721' and`point`='80';
+DELETE FROM `creature_movement` WHERE `id`='79721' and`point`='81';
+DELETE FROM `creature_movement` WHERE `id`='79721' and`point`='82';
+INSERT INTO `creature_groups` (`leaderGUID`, `memberGUID`, `dist`, `angle`, `flags`) VALUES ('79720', '79721', '7', '3.14159', '1');
+-- They are running
+UPDATE `creature_movement` SET `script_id`='11' WHERE `id`='79720' and`point`='1';
+-- Donna wants her dolly
+ALTER TABLE `creature_movement_scripts` 
+ADD PRIMARY KEY (`id`);
```
brotalnia: And what's the deal with that? You are supposed to be able to add multiple commands to a script, otherwise you wouldn't be able to make very sophisticated scripts, would you? The script tables have no key for a reason!

sipertruk: Yep forgot to remove that.

sipertruk:
I use to have warnings at server launch when my db missed migrations.. not anymore.

PR #1293, opened 2017-06-30
Manaweave Robe (ID: 7509) corrections

some2:
Manaweave Robe (ID: 7509):

BuyPrice/SellPrice should be 1376/275 instead of 2769/553
ItemLevel should be 15, not 20

Source:
http://www0.xup.in/exec/ximg.php?fid=60425525 (BradyGames Official Strategy Guide)
https://web.archive.org/web/20050913152711/http://wow.allakhazam.com:80/db/item.html?witem=7509


StadenElysium:
Research indicates that this was buffed (the current stats) with TBC when they added it as a reward to 'A Simple Robe'.

PR #1321, opened 2017-07-02
Reins of the Winterspring Frostsaber (ID: 13086) corrections

some2:
Reins of the Winterspring Frostsaber (ID: 13086)

BuyPrice should be 1000g (900g /w exalted bonus), not 100g
ItemLevel and RequiredLevel should be 60, not 40

Sources:
https://web.archive.org/web/20051128191751/http://wow.allakhazam.com:80/db/item.html?witem=13086
https://web.archive.org/web/20070312050430/http://www.thottbot.com:80/i13086 (See comments, eg by Sheesh)


StadenElysium:
This is true, up until 1.12.1. However, we're using the 1.12.1 mount system despite not being on that patch. 

I'll leave this open since we're reworking the itemisation and should be able to fit it in somewhere.

Phatcat:
better sources; http://blue.mmo-champion.com/topic/39485-patch-1121-riding-skill-changes/
http://blue.cardplace.com/newcache/us/28355274.htm

PR #1322, opened 2017-07-02
Beheading Blade (ID: 15253) corrections

some2:
Beheading Blade (ID: 15253)

RequiredLevel should be 57, not 47

Sources:
https://web.archive.org/web/20050404061627/http://wow.allakhazam.com:80/db/item.html?witem=15253
http://www0.xup.in/exec/ximg.php?fid=11454456 (BradyGames Official Strategy Guide


BBCnewstoday:
are you joking? 40 dps item req lvl57? Any other lvl 57 two hand item is 48dps.  Just useless "fix"

StadenElysium:
It might be a minor fix but it's not useless. All corrections are appreciated.

PR #1323, opened 2017-07-02
Evergreen Pouch (ID: 11020) corrections

some2:
Evergreen Pouch (ID: 11020)

Set RequiredLevel to 0 instead of 47.

This were at 0 in vanilla. Nostralius changed this to 47. I dont know why, see
https://forum.nostalrius.org/viewtopic.php?f=2&t=31252

### DIFF
--- a/sql/migrations/20170701194500_world.sql
+++ b/sql/migrations/20170701194500_world.sql
@@ -1,0 +1,4 @@
+INSERT INTO `migrations` VALUES ('20170701194500'); 
+
+-- Set RequiredLevel for Evergreen Pouch (ID: 11020) to 57:
+UPDATE item_template SET RequiredLevel = 0 WHERE entry = 11020;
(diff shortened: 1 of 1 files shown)


BBCnewstoday:
They changed it to prevent alt-using it=farming

druidish:
As BBCnewstoday said, this was a half baked fix that Nost did to prevent alts from farming morrowgrain.

If this commit is going to get pulled then someone needs to fix the root issue which is the following npcs shouldn't sell items to players that haven't completed the Un'Goro Soil quest.

Vendors
http://db.vanillagaming.org/?item=11020#sold-by

Horde Quest
http://db.vanillagaming.org/?quest=3761

Alliance Quest
http://db.vanillagaming.org/?quest=3764

brotalnia:
I think i fixed Mathrengyl Bearwalker in the class trainers gossip fix, but according to the mangoszero database he should be selling items after completing Morrowgrain Research 2, not Un'Goro Soil.

http://db.vanillagaming.org/?quest=3785

druidish:
If that's that case you'll need to tweak that so they've either completed Morrowgrain Research 2 or are currently on it. Or better yet link it to Morrowgrain Research 1 as that's when the player will get their initial seeds. If you link it to Morrowgrain Research 2 completed then there's a small chance that a player might not get enough morrowgrain to complete it.

thepparker:
Resolved in the new progression system

PR #1341, opened 2017-07-04
Hurricane (ID: 2824) corrections

some2:
Hurricane (ID: 2824)

BuyPrice/SellPrice should be 160159/32031 instead of 160160/32032

I found a lot of items with one copper price difference. Any idea why?

Source:
http://www0.xup.in/exec/ximg.php?fid=66640670 (BradyGames Official Strategy Guide)
https://web.archive.org/web/20050227011012/http://wow.allakhazam.com:80/db/item.html?witem=2824


some2:
This have been changed in patch 1.11.0 to 160160/32032. So the current db entry is correct.

PR #1381, opened 2017-07-07
Reset EventAI phase at grid unload/stop

sipertruk:
On grid stop or unload, creature's AI need its phase reset which happens on evade.
Not resetting the phase lead to unexpected problems because a lot of ai scripts (403) are incrementing it on aggro instead of affecting it. 
So after grid restart the AI is at any phase it was before, on aggro it's incremented from there.


StadenElysium:
This was changed by Nostalrius as a hackfix for an issue with Nefarian resetting during P1, as far as I know. Leaving this note as a reminder that the encounter will need testing.

sipertruk:
I couldn't find a problem with nefa, though I don't know how that bug triggers exactly.

StadenElysium:
As far as I know, it could be triggered by having a hunter feign death between P1 wave spawns. It was an issue for a couple of days on Elysium too, not too long after launch.

sipertruk:
I couldn't trigger any bug with feign death, nor vanish. As long as something else is in his threat list he switch to it.

brotalnia:
@sipertruk A recent commit introduced a shutdown crash on Windows builds, and i think it could be this one. Since the code you changed seems to be called on shutdown, do you think the fault could lie here? Maybe the AI is already deleted when you try to call EnterEvadeMode. Do you think that could cause a crash on shutdown?

The crash does not happen with the last build i used from two weeks ago. https://github.com/elysium-project/server/commit/34e12795f78b7cc8140df17d94143545b4c7605c So it must be one of the commits after that one which introduced it.

I put a printf before and after both lines you changed.
crash2

And here is what i see in my console at shutdown.
crash1

As you can see, it never even prints the message from the line after it. What do you think?

sipertruk:
Some AI is teleporting in Reset(), as the map is unloading it fails and it calls CreatureRespawnRelocation, creating a loop. Fix on the way.

Issue #982, opened 2017-06-05
The Missing Diplomat (Dustwallow Portion)


Phoosy:
The teleportation scripts that are supposed to move/clone Archmage Tervosh (NPC 4967, Quest IDs 1265, 1267) and Lady Jaina Proudmoore (NPC 4968, Quest ID 1267) temporarily to the player's location on completing certain parts of the chain don't fire. In particular:

Quest ID 1265 should 'teleport' Tervosh to the tower outpost (unclear about specific positioning)
Quest ID 1267 should 'teleport' Tervosh AND Jaina nearby Private Hendel (NPC ID 4966), invoke a quick conversation, and allow you to speak to Jaina to end the quest chain.

Instead, the quest requires players to run all the way back to Dustwallow, talk to the NPC in their normal spot, then run all the way back out again for each step.

QAkvastskaft:
WIP here: https://github.com/elysium-project/server/pull/2715

PR #1395, opened 2017-07-07
Missing Diplomat rewrite

kampeador:
Summary:
- Full quest event implementation (Missing Diplomat id:1324). https://github.com/elysium-project/server/issues/982
- Full quest event implementation (Missing Diplomat id:1447).
- Full quest event implementation (Missing Diplomat id:1249). 
- Rewritten db side implementation(The Missing Diplomat id: 1267).


Phoosy:
A PR of this nature would address https://github.com/elysium-project/server/issues/982

src/game/Objects/Creature.cpp
```diff
@@ -948,6 +948,10 @@ bool Creature::AIM_Initialize()
     CreatureAI * oldAI = i_AI;
     i_AI = FactorySelector::selectAI(this);
     delete oldAI;
+
+    // Handle Spawned Events, also calls Reset()
+    i_AI->JustRespawned();
```
sipertruk: AIM_Initialize is not only called on spawn. Instead you could, for example, do what i did here to call JustReachHome just after spawn : 

~~~~c++
    bool bJustCreated;

    boss_archaedasAI(Creature* pCreature) : ScriptedAI(pCreature)
    {
        ....
        bJustCreated = true;
    }
    void JustReachedHome()
    { }

    void UpdateAI(const uint32 uiDiff)
    {
        if (bJustCreated)
        {
            bJustCreated = false;
            JustReachedHome();
        }
        ....
    }
~~~~
https://github.com/elysium-project/server/blob/development/src/scripts/eastern_kingdoms/uldaman/boss_archaedas.cpp#L127

kampeador: I agree with you here. I've rushed a little bit with that change, by trusting original ManGOS source code too much and all my latest scripts are already using your approach.
Its a good alternative, but i don't like it from performance perspective. There can be alot of NPCs, that will use similar trick plus some scripts require them to be in active state and preventing map cell unloaded for some time. If compiler generates dynamic branch here, it will be a performance loss for nothing. Init function should be called only once.
If not this one, the other more effective alternative should be implemented in the future. For example, ZoneScripts already have OnCreatureCreate method.

All scripts are ready, and I'll update this PR soon. There will be a lot of of changes. I'll probably make a video to showcase them.

kampeador:
All features are implemented. The only thing i left are factions. For example: "neutral faction id 189". I took it from the deer in Ashenvale. Visually, it works fine. Blizzlike number will probably be 34, but db has alot of bugs, factions are messed up, resulting in collisions. That will require a ton of changes in DB and a lot of testing, so i left it as it is now.

thepparker:
You've added an NPC with entry 25000 for the cell trigger, but then later summon NPC 20102 with
`m_cellTrigger = m_creature->SummonCreature(20102, -3753.458008f, -4455.059570f, 64.975151f, 0.0f, TEMPSUMMON_MANUAL_DESPAWN, 0, true);`
instead. I assume this is unintentional. If it's not, let me know. Otherwise, I'm ready to merge it.

kampeador:
20102 and 25000 are identical, but i've added my own trigger entry(25000) to eliminate possible collisions with other scripts and databases.

thepparker:
Reported by QA:

> #1395:
Tervosh did not teleport to the player when http://web.archive.org/web/20070211044933/http://www.thottbot.com:80/?qu=1265 was triggered complete
Jaina's party stays (maybe) too long a time after the Hendel scene is played out. currently ~2min. Otherwise chain is easily completable

kampeador:
> Tervosh did not teleport to the player when http://web.archive.org/web/20070211044933/http://www.thottbot.com:80/?qu=1265 was triggered complete

All features were implemented for this PR. Quest event for id:1265 is not listed in change log.
I've completed it 3 times on retail and 2 of 3 were not smooth. I haven't found a sniff dump from the 3 attempt yet.

> Jaina's party stays (maybe) too long a time after the Hendel scene is played out. currently ~2min. Otherwise chain is easily completable

All timers are blizzlike. Once event is completed, Jaina's party will stay for 90 seconds in that area.
That's why i've added this line in my script:
`TELEPORT_BACK_DELAY = 90000`
Value is in milliseconds. It can be tweaked depending on the server's population, but keep in mind that some players read quest texts :)

This quest chain was very complex compared to other quests in the game. There is no phasing/instance and area is contested. Blizzard designed it like a single player/group scenario. There are a lot of design flaws with high population, where everything can happen with everything at any time. 
For example:
Tervosh is our main quest giver. He is not available in the tower during quests 1265 and 1324.
1 min during 1265
2 mins during 1324
There can be only 1 Tervosh at a time, so all quests have collisions with each other. If **player 1** starts 1324, then **player 2** who starts 1265 will have to wait for him and vice versa. I am thinking about adding an option to disable this feature on the server with high population, where Tervosh, Jaina and Pained will not disappear from the tower . What do you think about this?

As for event:1265, I'll make a separate PR for it later.

kampeador:
Don't push this version to live servers yet. I've returned to this quest chain again and found some new issues. I'll release a huge upgrade soon. 100+ changes. Some events were changed or completely rewritten and behave differently.

Issue #802, opened 2017-05-20
Pally Judgement debuffs stacking exploit.


Imbaslap:
Currently on all Elysium servers, All paladin judgement debuffs that are applied to a mob can stack with multiple ranks of spells. This behavior goes against the whole "debuff rank" system implemented by blizzard where a stronger spell or debuff will take precedence over a weaker one, and no multiple rank buffs/debuffs/hots/dots can stack with eachother.

Correct Behavior: When using a lower rank, it should report it and "and more powerful spell is already active", using the same or higher order to overwrite the original spell on the target.
It is the same in all judgments with rank, so even the Judgment of the Crusader, Judgment of Wisdom.

https://forum.nostalrius.org/viewtopic.php?f=43&t=17451 was a long discussion years ago but was considered a bug and lack of evidence.

http://wow.allakhazam.com/forum.html?forum=243&mid=11287757594799940 

after reading through the same source linked. this was reported fixed around patch 1.8 (see below snip)

> #7Oct 13 2005 at 12:31 PMRating: Decent
> Cathela
> Scholar
> **
> 961 posts
> Quote:
> Multiple paladins can stack different ranks of the judgements on the same mob. The last time my raid group killed Ragnaros we didn't have enough priests to do AoE heals on the rogues, so the three paladins in the melee groups coordinated different ranks of Judgement of Light on him, and the rogues and dps warriors ended up healing themselves a lot.
> 
> Reportedly this was nerfed in the 1.8 patch so that now you can only have one rank each of JoW and JoL on a monster, though I haven't had a chance to confirm this. 
> 
> Also, there's apparently some kind of aggro screwup when JoL and JoW proc since the patch. Not sure if this has been hotfixed or not. 

If paladins were "supposed" to be able to stack multiple ranks of judgements on a mob, then the Loatheb encounter in Naxxramas would showcase these results during patch 1.11.

after searching through multiple videos, no guild or raid had multiple rank judgement debuffs on the boss post patch 1.8 (when the nerf/fix was implemented by the above source).
https://www.youtube.com/watch?v=kF-rAvoOV7k
https://youtu.be/MW2uIvXVeb0
^ some examples.

I recently saw a stream of a guild abusing this bug on Viscidus in AQ40 to outheal the nature damage volley and dot placed on the raid.  This sort of bug will cause problems in current content and future content if left in its current state, as well as granting a large advantage to the Alliance faction on Elysium Project.

theloras:
[...]

PR #1423, opened 2017-07-10
Fix: Multiple Ranks of JoW/JoL on single target

Schwarzschildboy:
I am unsure if this one line will fix the issue entirely, but I added this line (and others to different branches) to hopefully fix the current issue with the Paladin's Judgement ability. Currently, Paladins are the only class able to stack multiple different ranks of the same ability on a single target, e.g. three different paladins in the same group can judge a single boss with three separate ranks of Seal of Light/Seal of Wisdom. This not only has obvious consequences for imbalanced gameplay, but also defies Blizzard's apparent rules regarding this game mechanic: abilities like Sunder Armor, Judgement of Light, Judgement of Wisdom, Curse of Weakness, etc. should not be able to stack on a single target, AND furthermore different ranks of these abilities should also not be able to stack on a single target.

### DIFF
--- a/src/game/Spells/SpellMgr.cpp
+++ b/src/game/Spells/SpellMgr.cpp
@@ -596,6 +596,7 @@ bool IsSingleFromSpellSpecificSpellRanksPerTarget(SpellSpecific spellSpec1, Spel
         case SPELL_AURA:
         case SPELL_CURSE:
         case SPELL_ASPECT:
+        case SPELL_JUDGEMENT:
             return spellSpec1 == spellSpec2;
         default:
             return false;


Schwarzschildboy:
#802 hashes this out after some digression, but any suggestions on where the issue lies would be appreciated!

Phoosy:
You may find some use in seeing how it was done in cmangos:
https://github.com/cmangos/mangos-classic/commit/0206d01e4c31ae1a0795c191f60615bb25829277
>* Paladin's Judgements (such as Judgement of Light or Judgement of Wisdom) are no longer multi-applicable on the same target.

Mostprovidedlinks:
you can't stack 2 of the same types, though you can put on judgement of light and wisdom if you have multiple paladins in the group.

Written in 2008

Mostprovidedlinks:
A 2.0 video shows you can stack more than one of thesame judgement however.

Mostprovidedlinks:
So if there are 3 paladins in the group we should be able to have judgements of wisdom light and crusader on the boss.

gearselysium:
According to patch 1.4:
Seal of Wisdom and Seal of Light should now replace each other properly as well as stack properly when used by different paladins.
Patch 1.10:
Blessing of Light - Different ranks of this spell were stacking without causing additional effect. It was never intended for them to be able to stack, so they can no longer stack.
So from this it seems clear that each paladin should be able to judge 1 instance, and that the rank stacking was because you could seal and judge different ranks by a single paladin, but the highest judged rank was the only one having an effect(edited)

PR #1449, opened 2017-07-12
Fix eyes of the beast

sipertruk:
PetAI::UpdateAI was running meanwhile.
Moved the pet return code in CMSG_SET_ACTIVE_MOVER handler so there's no need to ignore relocations.
Now in stay mode it's not returning as intended.
Set the move mode (run/walk) as identical as the hunter, instead of walk by default.


sipertruk:
The return handling needs improvement, wip.

PR #1450, opened 2017-07-12
[Naxx] Patchwerk major abuse fix

BBCnewstoday:
Hateful strike won't be used if highest HP player is standing behind Patchwerk. There is blizzlike way to fix this major abuse on this boss.


brotalnia:
I think those scripts are simply placeholders. They seem to be working on naxx in another secret repository. No point in submitting fixes here until they merge the new scripts with the development branch, as they have been most likely completely rewritten.

gemt:
The scripts available here are indeed not the ones used on ptr, but this was not fixed anyway so thanks for the heads up!

PR #1521, opened 2017-07-16
CompareAuraRanks should compare more effects

sipertruk:
Some spells have ranks with the same base value in the first effects.
Example Frost Shock :
image


thepparker:
What does this achieve?

sipertruk:
Better comparison between aura ranks every time this function is used. 6 times in the core currently.

sipertruk:
If you need a real example :
PlayerControlledAI is building a list of the player's abilities with only highest ranks, lower ranks are left in that list because of that bug.

thepparker:
Cool, was just wondering what it addressed

sipertruk:
But then I just realize it would have side effects in the aura incredible machine.
Nevermind PlayerControlledAI needs more something like CompareSpellRanks.

PR #1567, opened 2017-07-19
fix Warsong Gulch flags respawn

WobLight:
`Add` is not called when respawn time is set to 0.


WobLight:
I've just noticed this fixes also regression with powerups in AB

Issue #1512, opened 2017-07-16
Warrior cannot execute at 20% HP


ELouisYoung:
The bug is: execute is only usable when a mob is 19% or less.
The spell clearly states "usable on enemies that have 20% or less health".
A warrior player's UI should show "20%" hp remaining and be able to use execute.

luigimannoni:
Amazing to have spotted this, fix was easy enough. PR pushed and waiting for review.

ELouisYoung:
@luigimannoni thanks so much for fixing this.  I be you would also be able to fix this equally as easy: https://github.com/elysium-project/server/issues/2033

luigimannoni:
@thepparker please keep this issue open, the PR fixes only the Sweeping Strikes proc

GuybrushGit:
https://youtu.be/sW4vbSJgF98?t=242

Execute only lights up at 19%

elysium-asura:
A TBC example in addition to this - https://youtu.be/TTTfA4sya90?t=255

After Discussion with @GuybrushGit, and video footage from both TBC and Vanilla warriors, it is decided that this is working "as intended".

PR #1582, opened 2017-07-19
Sweeping strikes Warrior execute can proc now to other targets at 20% or less

luigimannoni:
Swapped mathematical comparison on player target on Warrior execute aura as execute can be cast when the target is exactly 20%.

Reported here: #1512


thepparker:
That's the logic for sweeping strikes procs on execute targets, not whether execute is usable on the main target or not. But yea, it should probably be <= 20% too

luigimannoni:
Update: The initial PR was supposed to fix the Warrior execute direct spell however this PR fixes the SS procs on secondary targets when the Warrior casts Execute on the main target. I edited the title for better clarification

PR #1585, opened 2017-07-20
Eranikus should whisper players who use his trinket

brotalnia:
Every time the trinket Chained Essence of Eranikus is used, the player should get a whisper from Eranikus the Chained.

http://db.vanillagaming.org/?item=10455


Ripper007:
Finally! I used to keep the trinket on retail only to hear those emotes.

PR #1595, opened 2017-07-20
fix TBB_ROOT_DIR use $ENV

dulingzhi:


WobLight:
`TBB_ROOT_DIR` is not supposed to be an ENV variable, just set it with CMAKE or use `TBB_INSTALL_DIR` instead

thepparker:
As Woblight said, use TBB_ROOT or the other alternative env variables

PR #1650, opened 2017-07-22
Sleeping creatures

kampeador:
I've heard, that Kronos has similar feature, so i decided implement my version.
- At at night time (between 9pm-5am) some creatures stop moving and go to sleep. Only bears and cats are affected, because they have a "sleeping" animation.
- Creatures are awake for some time after combat/respawn.
- Aggro range is the same, so it doesn't affect gameplay.
- It's not a blizzlike feature. There were some creatures like similar to this on retail, but i have no verified data.
- ManGOS had a similar script, but it was unfinished and inefficient in terms of performance.
- All creatures were tested to ensure, that there are no collision with other scripts.
There are 10 creatures that use buggy data with generic_spell_ai, in which Broken Tooth has fireball spells and other crazy things. This script overwrites them.


brotalnia:
Why add custom features? Elysium is a blizzlike server. :confused: 

I would move the script to scripts/custom/ since that would seem like the appropriate place, and instead of a migration put the sql part in sql/custom/. That way people who use the core can enable it if they want, but it doesn't get added to live realms. Many people do not enjoy seeing custom content.

luigimannoni:
Actually this is the behaviour on a blizzlike server some mobs do sleep, so it isn't exactly a custom feature.
The question is, this was a thing in vanilla or was it introduced later? Can't find anything on internet.

brotalnia:
Best source i can think of would be Joana's leveling videos which are form patch 1.9 i believe. It is multiple hours long and some of it is during the night. You can tell by the moon icon at the top right corner.

Here she is killing panthers in STV, and they don't seem to be sleeping.
https://youtu.be/0ui-onkUGGQ?t=4m6s

kampeador:
**brotalnia**
The reason i didn't put this in custom, because script behavior itself is blizzlike, but its data is custom/unverifed.
These are verifed, but they are independent from game time:
https://ibb.co/jxUmkQ
If someone has some info about this, i can update this script, if not, i'll close this PR and move to other subjects.

This script was partly derived from my blizzlike implementation for quest "The Sleeper Has Awakened" i currently working on. Its complexity currently is like my full Missing Diplomat scripts together.
I've took some break and spend about 40 mins to implement this Kronos-like "ultimate" feature. I've also experimented with "smart" AI behavior similar to theirs which uses dynamic movement.

This feature may seem too custom for Elysium, but what do you think about others?
For example: changing profession specializations, switch from Tribal to Elemental leatherworking, etc. Currently server uses a non-existing spells from BC. Those scripts need to be disabled completely or rewritten.

luigimannoni:
So it is a case of spawning a monster/npc in sleeping state rather than having one go to sleep?

kampeador:
Yes.

Issue #1467, opened 2017-07-13
Resupplying the Excavation


Phatcat:
quest not properly implemented;

completed on pickup (shouldn't be)
dwarves are not fighting when you find them (they should be)
the dwarves that ambush you should be ambushers, not raiders.

ref; https://github.com/cmangos/classic-db/issues/71
https://github.com/cmangos/mangos-classic/commit/3743bf7b02c4f7e812dc3af5f95b895bbfb348e8
https://github.com/cmangos/classic-db/commit/aa237c43ebf5260fd7ba463f9ae1cbe2bd1b617b

kampeador:
Thanks for the links, but i will probably implement my own version of this event.

PR #1671, opened 2017-07-23
Resupplying the excavation

kampeador:
Quest event implementation for (Resupplying the Excavation id:273).
Unlike other scripts, my version handles rare cases, which can occur with high population. https://github.com/elysium-project/server/issues/1467


kampeador:
The Travis CI build info seems wrong. Script is well tested and compiles fine.

PR #1674, opened 2017-07-23
Prevent creatures using a MC spell on a single player

SpaceVector:
Added support for creatures that have a spell that allows to mind control the player. What happened was that the player wasn't hostile anymore and was removed from the threat list of all creatures that had the player on their threat list.

A Blizzlike behavior is that the creature won't cast any MC spell on a player that is alone.

To fix this, we add a new flag to CastFlags that we check whenever a creature will cast a spell. We can now set this flag for any creature's AI and fix this issue.

In this commit, I only fixed Salia. But there might be other creatures that share this problem, and we must update their CastFlags in their AI too, which is now made easy with this new flag.


sipertruk:
You could just use TARGET_T_HOSTILE_RANDOM_NOT_TOP.

SpaceVector:
That would make the creature not cast its spell on a totally random target since a player holding the highest threat would never get it. That would be a hack fix to me. I saw no other choice than defining a new flag for cast flags to support these spells.

thepparker:
This could potentially be made much more generic by simply checking if the spell being casted applies SPELL_AURA_MOD_POSSESS or SPELL_AURA_MOD_CHARM

SpaceVector:
I first thought about that, but was worried we would want the same behavior for another kind of spell. Went that route instead, but it requires to update the mobs ai in the db everytime we face this (rare) issue

javierhimura:
There was at least one NPC who could mind control a solo player in Vanilla.
http://wowwiki.wikia.com/wiki/Dr._Weavil

"Note that he cannot be soloed by any class without a combat pet, and the only classes able to get away from an endless loop of mind control are: Shaman, Hunter, Warlock, Paladin, and Mage, as Weavil will reset and return to full health when all opponents are dead or mind controlled. (See Tips for methods to escape). "

Of course it is impossible to solo him with a level 60 player but still technically this NPC never has any mechanic to avoid MC a solo player.

SpaceVector:
Dr Weavil's ability to mind control solo players could be a simple oversight in retail, as they didn't give it the right aura maybe. I don't know if we would want that blizzlike behavior here though. Anyway, if there  is any suspicion that there are other spells that are not casted on a single player (not necessarily a MC spell, could be anything), the flag solution should be kept over simply checking these auras

PR #1696, opened 2017-07-24
Fix Cortello's Riddle quest

SpaceVector:
The last part of the series have you interact with a chest in the Hinterlands, which was impossible because a wrong quest ID was set. Fixes issue https://elysium-project.org/bugtracker/issue/2490


thepparker:
This parameter doesn't actually appear to be used for anything, and I am able to complete the quest even on the old value of data1. I think the reporter of that issue was trying to obtain the quest from the box or something, rather than hand it in there.

SpaceVector:
It didn't work on live servers when I got to do this quest. I'm 100% sure. Just fixed the quest Id (data1) without trying the previous one indeed, and assumed it fixed the issue.

sql/migrations/20170724123610_world.sql
```diff
@@ -0,0 +1,4 @@
+INSERT INTO `migrations` VALUES ('20170724123610');
+
+-- Fix quest ID for Cortello's Treasure
+UPDATE `gameobject_template` SET `data1` = 626 WHERE `entry` = 2556;
```
ChehST: What is '626' and '2556'? Where did you give this values?
p.s. just intrested, optionaly to answer :)

ChehST: ok, '626' is quest's id, what is '2556'?

brotalnia: the id of the gameobject

brotalnia:
@SpaceVector I just tested this quest chain on the latest revision. I was able to accept and complete all 3 quests, including the final one that has to be turned in at the underwater chest. Therefore i am going close this pull request, as everything relating to the quest is already working. Most likely it was intentionally or unintentionally left disabled on some of the live servers at the time you had tried to complete it.

PR #1708, opened 2017-07-24
Fix compile errors & style in 7b31cf3559cb1c75e0c370e223e48b26aa75c0aa

EinBaum:
- error: cannot pass objects of non-trivially-copyable type ‘std::string {aka class std::basic_string<char>}’ through ‘...’
- Tabs instead of spaces


EinBaum:
https://github.com/EinBaum/server/commit/7b31cf3559cb1c75e0c370e223e48b26aa75c0aa

PR #1711, opened 2017-07-25
Fix player log of honor gain for AV events

sipertruk:
You're told that you died and you're awarded points for it :
capture2

Patched :
capture1


brotalnia:
Would this fix the same issue when turning in bg marks? It tells you that you've got a honorable kill on yourself.

sipertruk:
I suppose all added honor points without a kill is inside that case, didn't really test though.

Issue #1755, opened 2017-07-26
weapon normalization


stesve:
http://wowwiki.wikia.com/wiki/Normalization

> Normalization refers to an Attack Power adjustment made to instant attacks in patch 1.8. Hunter ranged instant attacks were also normalized in patch 1.10. 

Functions to handle normalization is already in the code, but every call to them has "normalized = false". 

Normalization should be ingame.

stesve:
StatSystem.cpp:
void Player::CalculateMinMaxDamage(WeaponAttackType attType, bool normalized, float& min_damage, float& max_damage, uint8 index)

.....

void Player::UpdateDamagePhysical(WeaponAttackType attType)
{
...
CalculateMinMaxDamage(attType, false, mindamage, maxdamage);



Unit.cpp:
void Unit::CalculateMeleeDamage(Unit *pVictim, uint32 damage, CalcDamageInfo *damageInfo, WeaponAttackType attackType)
...
        subDamage->damage = CalculateDamage(damageInfo->attackType, false, i);

PR #1768, opened 2017-07-27
Weapon normalization

stesve:
#1755 - activates weapon normalization.


thepparker:
This is wrong, and incorrectly normalizes all melee damage such as white hits and 'next attack' abilities. Attack speed normalization is applied in EffectWeaponDmg here https://github.com/elysium-project/server/blob/development/src/game/Spells/SpellEffects.cpp#L3928, and only affects certain abilities, i.e. those with SPELL_EFFECT_NORMALIZED_WEAPON_DMG

stesve:
Thanks for the info.

PR #1789, opened 2017-07-27
Crash fix attempt

sipertruk:
It crashes here on live servers https://github.com/elysium-project/server/blob/development/src/game/Maps/Map.cpp#L4061 no other information.
Attempt to smash the problem for the case it's caused by ritual objects.


sipertruk:
So what fixed this crash ?

thepparker:
This is apparently live

PR #1836, opened 2017-07-30
MacOS compatible changes incl. Travis build

martinjlowm:
Adds an extra Travis CI build for MacOS and applies MacOS compatible
changes to enable compilation thereon.

- MacOS assert macros conflict with some functions and is therefore
  turned off.

- Works around a uint64-redefinition issue

- Adds additional search paths for OpenSSL

- Assures c_str() conversion of C++ strings when using PSendSysMessage

- Uses size_t for storing thread identifiers. POSIX threads on MacOS are
  identified as pointers, casting these to size_t uses the memory
  address as identifier. This change is temporary, for as long as ACE is
  used.


EinBaum:
Not compatible with https://github.com/elysium-project/server/pull/1708

martinjlowm:
I've removed changes that were also included in #1422 and #1708 and this pull request now depends that both of these are applied first.

PR #1944, opened 2017-08-05
Implement Quest Greetings.

brotalnia:
This implements quest greetings. These are texts that are shown when an NPC is offering multiple quests. It looks like gossip, but it is not technically gossip. The NPCs who use a quest greeting have no gossip menu, nor do they have the gossip npcflag, nor does the text exist in the `npc_texts` table where gossips are stored. The only other core that has this implemented is TrinityCore's master branch, but this is a feature that has always existed in WoW.

Here is an example of a quest greeting:
https://youtu.be/Jgcn0jgdyvE?t=2m24s
Notice how the text is shown only when there are two quests to choose from.

### DIFF
--- a/sql/migrations/20170805164518_world.sql
+++ b/sql/migrations/20170805164518_world.sql
@@ -1,0 +1,82 @@
+INSERT INTO `migrations` VALUES ('20170805164518'); 
+
+-- Structure for table quest_greeting
+CREATE TABLE IF NOT EXISTS `quest_greeting` (
+  `entry` mediumint(8) unsigned NOT NULL DEFAULT '0',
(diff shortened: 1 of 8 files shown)


henkethebenke:
What about the "quest greeting" at [2m39s](https://www.youtube.com/watch?v=Jgcn0jgdyvE&feature=youtu.be&t=2m39s) when there is only one quest to accept? Or is that something else?

brotalnia:
It's not a quest greeting if it appears when only 1 quest is available. That creature has gossip. The creatures who use a quest greeting don't have the gossip npcflag, so when only 1 quest is available, it jumps straight to the quest if you talk to them.

Issue #1529, opened 2017-07-17
Rallying Cry of the Dragonslayer missing Internal CD (onyxia/nefarian head)


knasen88:
> The server or game patch that your issue pertains to (e.g. Anathema, patch 1.10).
Elysium, 1.4 and 1.6 (onyxia head and nef head)

> A clear, concise description of the current behaviour.
Currently, you can turn in the buff back to back even if there is a head up in the chains/on the pole and recieve the buff

> A description of the expected behaviour.
When placed upon the spike/chain, it should not fire any more new buffs until the previous has despawned after X hours, whereas X is uncertain but believed to be around 6 hours (same as WCB 21600(000))
This would mean 2 buff opportunities per X hour (currently only 1 on elysium but Nefarian is soon added)
I suggest copying the WCB script and mimmicking it to work with the heads somehow.

> Where applicable, provide relevant sources that show or describe the correct behaviour.
Patch 1.7 also mentions it briefly that to buff, there can be no (same) head hanging.
https://web.archive.org/web/20070602172347/http://www.worldofwarcraft.com/patchnotes/patch-05-09-13.html

http://alt.games.warcraft.narkive.com/nbkesSJ1/turing-in-onyxia-head-turnin-city-buffs

> When I have handed-in Onyxia's head, nobody was buffed,
because a head already dangled from a stone ridge on the
entrance to Stormwind (valley of heroes). I don't know
how often it is removed, but I've heard that only one head
can be handed-in per week.
Not at all! It's a couple of hours normally (or is that 24h? can't
remember), check the guards to see what buff they have, this will let
you know if your head will buff everyone around.
AFAIK* you can get the buff only every six hours. But the buff (and the
head itself) are only two hours. But I think Nefarians head is on a
different counter; but you can't receive the buff more than once.

thepparker:
Having the 6 hour cooldown doesn't really make sense with a population the size of Elysium's to be honest. With over 50 guilds raiding per faction, that's a large number of Onyxia heads being turned in every day.

Not being able to acquire the buff because one of them was turned in a few hours before your raid (which will happen very often) leaves a bit of a sour taste.

fgsfdss:
Though that same argument applies for Warchief's Blessing, which also has said cooldown and in a bugged state at that (requires two of the quest to be turned in before the CD starts ticking).

thepparker:
I agree, and I don't think the blessing should have such a long cooldown either.

strudelkuchen:
[...]

PR #179, opened 2017-03-28
Implemented the Warchief's Blessing event

GuybrushGit:
https://elysium-project.org/bugtracker/issue/1200
Aditional sources:
https://www.reddit.com/r/wow/comments/1sob6t/people_which_only_play_one_faction_talk_to_me/cdzqh4s/
http://wowwiki.wikia.com/wiki/Herald_of_Thrall

Event behaviour:

When turning in the quest a Herald of Thrall is spawned on the Crossroads.
2 secs later he and Thrall yell _Honor your heroes! On this day, they have dealt a great blow against one of our most hated enemies! The false Warchief, Rend Blackhand, has fallen!_
10 secs later Thrall yells _Be bathed in my power! Drink in my might! Battle for the glory of the Horde!_ and the Herald yells _Be bathed in the power of the Warchief! Drink in his might! Battle for the glory of the Horde!_
2 secs later both cast the Warchief's Blessing.
The buff also goes out to the rest of Orgrimmar, Razor Hill and Camp Taraujo.

I could find no video recordings so the timers are an approximation.

PR #1978, opened 2017-08-07
warchief CD removed (same base script)

knasen88:


brotalnia:
1000 ms is 1 second. So you could probably start the event again before it has even finished with this change...

knasen88:
Made it 15 seconds, should be enough

thepparker:
Hmm, this is a bit of a contentious subject

knasen88:
#179 The original claim for 6 hours is total bullshit. REMOVE. 
#1529 has better sources for a CD and yet nothing is done.

GuybrushGit:
This should really be approved, makes no sense to have this cooldown based on some vague forum comments and no cd on the other buffs.

SiguDS:
Plz mercy let us not have to sit and camp this stupid buff just because of a 6 hour cd. that you won't implement (eventhough the evidence is there) on ony/nef head.

strudelkuchen:
The worst part is not even the 6h cd, but that it's not reliable. Many times we need to pop multiple heads before it even works, even though 6h+ has passed.

knasen88:
yes, I too believe there is something flawed. Some speculate the actual cooldown doesnt start until another quest is turned in, making the CD really unreliable because that may very well take some time. If gms could take their time and try it on their PTR or something

kvalitet:
This is really something that needs to be looked at!

There is 2 problems here, first off currently there are 2 different policies when it comes to these buffs: Onyxia/Nefarian has been decided to not be on a 6 hour blizzlike cooldown - but this one for some reason has? Either all needs to be on the cooldown or none should be.

Second, the 6 hour cooldown does not work at all! I have been told it is problary because of some systems that save memory(?) for the server so it cannot be fixed.. But even so - if you cannot fix it to a direct 6 hour cooldown - instead of this random 6-9 hour cooldown(that makes the buff useless), you need to remove the cooldown as this pull request does.

We need a solution here.

Issue #1934, opened 2017-08-05
Can't return the reward for Arathi Basin marks


Taurior:
The pnj doen't give me the quest to return the reward for Arathi Basin marks

Brunheda:
This is already open: #1930

Taurior:
so u can resolve the problem ?

thepparker:
Duplicate

Issue #1875, opened 2017-08-02
Returning WG mark of honor bug(no quest!!)


saze111231234:
I can't return my moh to npc.

He only accept AB moh. most of my WG moh is gonna disappear
![cap 2017-08-02 02-35-27-117](https://user-images.githubusercontent.com/30648650/28860845-3610a140-772c-11e7-8e14-b56e2dacdd1f.jpg)

Anathema, horde
ID : saze

Stackieks:
I have the same problem. I cannot obtain the qeust to return warsong gulch marks. I made a ticket and the GM told me to retake the qeust (I cant see the qeust in my qeustlog) and delete my WDB folder. This didnt solve the problem and he told me to go here.

berndbrotx:
I have the same problem. The npc only shows me the 'Invaders of Alterac Valley' quest. I can not turn my wsg or ab moh in. Also deleted my WDB folder but it didn't help.

bora994:
me too

Travnastington:
I am also having this issue.

thepparker:
Duplicate

Issue #1781, opened 2017-07-27
WSG tokens


Humppastara:
SERVER: Darrowshire (horde)

Several other people can turn in the Warsong gulch marks of honor.

Can't turn in WSG tokens at Horde Warbringer in orgrimmar or undercity and the WSG place in barrens didn't have anything either.
Does it require to be friendly with Warsong outriders? Currently neutral.
Tried turning them in on shaman and rogue. Nothing.

Phoosy:
This is a duplicate issue report, use the search function next time. Please close to reduce clutter.

thepparker:
Duplicate

Issue #1710, opened 2017-07-24
Freedom For All Creatures: Quest broken in Feralas.


CobaltEZ:
The quest Freedom For All Creatures is currently bugged for me and other players in Feralas. The quest appears on questie and the npc that gives the quest is there atm but the quest is not available through her. The lock on the dragon cage even says it requires the bamboo key (quest item) when moused over. This quest being unfinished also blocks the acceptance of the second part of the quest Doling Justice. 

Thank you and have a nice night/day Elysium team!

PR #1979, opened 2017-08-07
Re-enable quests left disabled after QUEST_METHOD_DISABLED implementation

### LINKED ISSUE #1940: Frostwolf Stable Master quest bug
The Frostwolf Stable Master in Alterac Valley is not giving the quest "Empty Stables" to harvest the wolves. Can you please fix this?

sipertruk:
Implement Quest Disable (#964) implemented the flag but some quests had already Method=1 in the database.

Correct me if I'm wrong but disabling quests should be done in itemization patches.

Involved quests :
image

Fix #1940 #1875 #1640 #1683 #1566 #1781 #1934 #1560 #1550  #1677 #1710 #1667


brotalnia:
You forgot to include quest 8429 in the query, even though it is listed in the screenshot.

PR #2015, opened 2017-08-09
Weapon skill fix part two point five

Cephel:
The fix was missing the ability miss chance calculation


src/game/Objects/Unit.cpp
```diff
@@ -3093,10 +3093,13 @@ float Unit::MeleeMissChanceCalc(const Unit *pVictim, WeaponAttackType attType) c
     if (!pVictim)
         return 0.0f;
 
-    // Base misschance 5.60%
-    float missChance = 5.60f;
+    float missChance = 5.60f; // The base chance to miss is 5.60%
+    if (pVictim->GetTypeId() == TYPEID_PLAYER)
+    {
+      missChance = 5.00f;  // The base chance to miss in PvP is 5%
```
thepparker: 4 space indentations

taurenrogue:
whys the miss chance base 5.6% instead of 5.0% (+defense-skill)
apart from no souces this is the only part triggering my ocd....

PR #2023, opened 2017-08-09
Implement .unstuck command.

brotalnia:
This implements a command to help players unstuck themselves without having to bother a gm.

- 1 Hour Cooldown.
- The user is afflicted with Resurrection Sickness.
- Cannot be used in combat, on flight paths, or in battlegrounds.
- All uses of the command are logged.

If used by an alive player it places a 10 second aura on them. When the aura expires, the player dies. If the player enters combat, casts any spells, takes damage, changes zones, or teleports while the buff is active, it is cancelled. Players may remove the aura themselves before the 10 second cooldown expires if they wish to cancel the effect, but the cooldown is still triggered.

When used by a dead player, it simply teleports them back to the graveyard.


Kittnz:
Wait, isn't this already implemented?

brotalnia:
Not to my knowledge? There is the .start command, but that just uses the auto-unstuck tool, which has a 10 second cast time and all it does is teleport you a few feet above your current position. It doesn't help when you are stuck falling under the world.

Kittnz:
Fair enough

PR #2054, opened 2017-08-11
Improvements to ticket system.

brotalnia:
Some changes to the ticket system that were suggested.

tickets

- Show category when a new ticket is created.
- Notify GMs when a ticket is edited or abandoned.
- Ability to list only tickets from a certain category.

The .ticket list and .ticket onlinelist commands will accept categories as an optional parameter.

The categories are: stuck, behavior/harassment, guild, item, environment/world, npc/creature/creep, quest/questnpc, technical, account/billing, character.

So if you wanted to see a list of tickets in the 'Stuck' category you would use:
`.ticket list stuck`

### DIFF
--- a/sql/migrations/20170811162753_world.sql
+++ b/sql/migrations/20170811162753_world.sql
@@ -1,0 +1,6 @@
+INSERT INTO `migrations` VALUES ('20170811162753'); 
+
+-- Edit texts for ticket system.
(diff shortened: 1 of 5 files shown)


Peter2345:
May be an idea to send account/billing to the forums, technical also???

brotalnia:
Actually it seems the client doesn't let you create tickets under account or technical. There is no button. So you won't see tickets from them, they are only technically valid since they exist in GMTicketCategory.dbc.

Peter2345:
I see one for account billing.

brotalnia:
no button
There is no button to create a ticket under this section.

Peter2345:
screen shot 2017-08-11 at 19 37 16

PR #2118, opened 2017-08-16
Implement new progression system.

brotalnia:
This implements a new progression system, making the separate sql files for every patch obsolete. We can now populate data for all patches in the same database, by just marking every entry with the patch in which it was added in the game.

It uses the WowPatch setting in the mangosd.conf file.
**Patch values go from 0 for 1.2, to 10 for 1.12.**

The following tables now have a `patch` column:
- `areatrigger_teleport`
- `battleground_template`
- `creature_addon`
- `creature_equip_template`
- `creature_equip_template_raw`
- `creature_template`
- `creature_template_addon`
- `item_template`
- `quest_template`

The `patch` column is a key and it indicates the patch in which this entry was added to the game. You can have multiple copies of every entry, but for different patches. The server will only load the entry for the latest patch that is not bigger than the current patch value set in the mangosd config file.

Example time:

If an item was added in patch 1.4, and had it's stats changed later in 1.10, you would need two entries in `item_template` for this item. 

One entry will have `patch` set to 2, this is the original version.
The other entry will have `patch` set to 8, this entry is for the updated version.

If the server patch value in the config is set to 0 or 1, this particular item will not be loaded at all and it will not exist. If the value is between 2 and 7, it will load the original version of the item that was added in patch 1.4. Once the config patch value is set to 8 or above, it will load the second entry, the updated version of the item.

**You only need to add additional entries for the patches where the item changed, not for every patch.** The server will know which one to load based on the WowPatch setting in the config file. I used the `item_template` table as an example, but this logic applies to all the tables i listed above.

The following tables now have a `patch_min` and `patch_max` columns:
- `creature`
- `game_event`

The `patch_min` column indicates the minimum patch that needs to be reached for this entry to work. The `patch_max` value indicated the maximum patch, after which it will not work. If the current patch value from the config file is outside of this range, the entry will not be active. They still get loaded to prevent startup errors, but they are set to disabled, so the creatures will not spawn, and the game events will not start.

**These  columns are not keys, so you can't have duplicated entries in these tables.** I've implemented these columns differently, because they are used differently than the template tables. For example, take the Purgation Isle issue. Up until 1.10 there are supposed to be different mobs spawned on the island, so we need to be able to specify a maximum patch after which the old mobs will not be spawned. And it does not make sense to have different creatures added with the same guid, as that would be confusing and could cause many mistakes. In the case of the `game_event` table, there are events that are only for specific patches, for example the AQ Gate Opening event, so we need a min and max patch values here too.

There is also a new table:
- `forbidden_items`

Structure:
```
`entry` mediumint(8) unsigned NOT NULL DEFAULT '0',
`patch` tinyint(3) unsigned NOT NULL DEFAULT '0',
`AfterOrBefore` tinyint(3) unsigned NOT NULL DEFAULT '0',
```

Items added to this table will not be sold by vendors or drop as loot.  The `AfterOrBefore` column indicates whether the item is forbidden after or before the patch specified. So if you want the item 17782 not to drop after we reach patch 1.5 you will add the following values:
`INSERT INTO forbidden_items VALUES (17782, 3, 0);`

Additionally, there is now a new condition type.
- CONDITION_WOW_PATCH (37)

The first value of the condition is the patch number, the second is 0 for equals, 1 for equals or above, 2 for equals or below. Conditions can be used for loot, gossip options, scripts and others.

Now it is up to people who work on progression/itemization to populate the correct patch values into the database. I will give a few more examples below to make things clearer.

### DIFF
--- a/sql/migrations/20170816025732_world.sql
+++ b/sql/migrations/20170816025732_world.sql
@@ -1,0 +1,57 @@
+INSERT INTO `migrations` VALUES ('20170816025732'); 
+
+-- Implement new progression system.
+ALTER TABLE `areatrigger_teleport`
+	ADD COLUMN `patch` TINYINT(3) UNSIGNED NOT NULL DEFAULT '0' AFTER `id`,
+	DROP PRIMARY KEY,
+	ADD PRIMARY KEY (`id`, `patch`);
+	
+ALTER TABLE `game_event`
+	ADD COLUMN `patch_min` TINYINT(3) UNSIGNED NOT NULL DEFAULT '0' AFTER `disabled`,
+	ADD COLUMN `patch_max` TINYINT(3) UNSIGNED NOT NULL DEFAULT '10' AFTER `patch_min`;
+	
+ALTER TABLE `creature`
+	ADD COLUMN `patch_min` TINYINT(3) UNSIGNED NOT NULL DEFAULT '0' AFTER `spawnFlags`,
+	ADD COLUMN `patch_max` TINYINT(3) UNSIGNED NOT NULL DEFAULT '10' AFTER `patch_min`;
+
+ALTER TABLE `creature_template`
+	ADD COLUMN `patch` TINYINT(3) UNSIGNED NOT NULL DEFAULT '0' AFTER `entry`,
+	DROP PRIMARY KEY,
+	ADD PRIMARY KEY (`entry`, `patch`);
+	
+ALTER TABLE `creature_addon`
+	ADD COLUMN `patch` TINYINT(3) UNSIGNED NOT NULL DEFAULT '0' AFTER `guid`,
+	DROP PRIMARY KEY,
+	ADD PRIMARY KEY (`guid`, `patch`);
+	
+ALTER TABLE `creature_template_addon`
+	ADD COLUMN `patch` TINYINT(3) UNSIGNED NOT NULL DEFAULT '0' AFTER `entry`,
+	DROP PRIMARY KEY,
+	ADD PRIMARY KEY (`entry`, `patch`);
+	
+ALTER TABLE `creature_equip_template`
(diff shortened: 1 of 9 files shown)


brotalnia:
If you want to make an item not drop for example, you have 4 different ways to do it with this system.

Using `forbidden_items`:
- Item will not drop or be sold by any NPCs.
- Item will still exist.
- You can query the server for it.
- Anyone who had it gets to keep it.
- You can remove it before or after a specific patch.

Using `patch` column in `item_template`:
- Item will not exist at all.
- Anyone who has it will have it removed permanently.
- You can't query the server for it.
- Can only be used to remove it before a specific patch, not after.

Changing `lootid` in `creature_template`:
- You can remove it from creatures with a specific creature id.
- Can also be used to change drop chance, since you specify another loot template.
- The item will still exist and can drop from other sources.
- You can query the server for it.
- Anyone who has it gets to keep it.
- You have complete control on which patches it drops, not just before or after, because you can make a separate `creature_template` entry and specify a different lootid for each patch if you wanted. You can have it drop one one patch, then not drop on the next, and then have it drop again on the patch after that.

Adding condition with type 37 to the row in the loot template table.
- You can remove it from a specific loot template.
- Conditions are evaluated at run time, so it impacts performance.
- Item still exists and can drop from other sources.
- You can query the server for it.
- Anyone who has it gets to keep it.
- You can have it not drop on a specific patch, after a patch, or before a patch.

As you can see, every method is suitable for a different situation. First we should populate the `patch` columns in the `*_template` tables with the patch in which they were first added to the game, and then we can handle items that were made unavailable at a later patch using one of the other methods. If they were removed from all sources, we add them to `forbidden_items`, otherwise we use one of the remaining 2 methods to remove them from specific creatures. If the creature's whole loot table changed, we add a new `creature_template` for that patch and change the `lootid`, other wise we can just add a condition to to the loot template entry.

PR #2130, opened 2017-08-16
Fix For The Fix To Fix Sunder Armor Procs

YoshGit:
Moved the Sunder Armor fix for triggering main hand weapon procs outside of a conditional statement that made it never execute.


thepparker:
It already procs where it is. The patch is not live yet.

PR #2146, opened 2017-08-18
Fix pool respawn (crystals, doll parts, herbs etc)

sipertruk:
What's happening is that when the same location is selected it won't respawn because already removed from the map's object structure if it's a GameObject.
For doll's parts it's very fast, for others things it ~~~'s very slow~~~ might be slower.


martinjlowm:
That is interesting. I had actually tried to see if this was the way it behaved. I wasn't able to reproduce it though.

But what these pools have in common is the spawn limit of 1.

sipertruk:
When limit > 1 the pool is fixed later when another object of the pool despawn.

martinjlowm:
Makes sense.

sipertruk:
Maybe it wasn't pushed to live till one month ago, I don't know. And not knowing is wasting time, beside #856 was lacking for GameObjects.

rickharrison:
Is the hypothesis for this fix that an object cannot respawn where it once was? I.E. for Black Lotus, it spawns in a zone until it picks a second spawn and then it does not appear meaning there are 0 lotus total?

sipertruk:
It's probably what you mean, let me try to be more clear about the initial problem.

If a different location is selected it spawns as usual, but if the same is selected Map::GetGameObject won't find it because it has been removed from the map already. (https://github.com/elysium-project/server/blob/development/src/game/Maps/Pool/PoolManager.cpp#L570)
If it's a single spawn pool there's now 0 object spawned.
Past that, the pool is not updated anymore because it only does when a player collects one.

~~
The function PoolGroup< GameObject >::ReSpawn1Object is useless.

thepparker:
You beat me to it :P

rickharrison:
@sipertruk @thepparker You mean that if a location where it was previously spawned is selected for the respawn? I'm not sure if this will fix it, because I have definitely seen Black Lotus spawn in the same location before it "bugs out" and no longer spawns at all.

For example yesterday, I was in an AV and saw Black Lotus respawn for the initial 3 hours (Initial spawn + 9 respawns). It spawned in the same location multiple times, but then just mysteriously stops respawning about 3 hours in.

martinjlowm:
That's how I understood it. I cannot say for sure I experienced it myself, but I haven't been able to reproduce the spawn issue locally. The doll- or Un'goro crystal issue are the best systems to test this because of their small pool size. 

I'm beginning to think that there must be a threading issue somewhere. For example, this issue and the warchief blessing not working consistently are two signs of something external that interferes. This is only speculation though, and I don't have a complete overview of the source code yet,  so I'm not sure if threading is the culprit.

rickharrison:
As another anecdote, I have gotten the same Black Lotus spawn in the same hour before and then hours later it presumably stops spawning all together.

Issue #1136, opened 2017-06-23
Judgement of command scales at 43% in stun but is only scaling at 23% normally (should scale 43% regardless of whether its a non stun or a stun)


Mostprovidedlinks:
https://www.youtube.com/watch?v=ay7RU0N-44Y

Judgement of Command hits in 390 spell damage gear (NO sanctity aura)  for exactly 342 or more damage on the rogue in this video, 4:95 and 384 damage on 10% berserking stance at 2:00
he appears to be hitting some wartriors for exactly 10% more damage
if you add all the healing enchants etc and gear he has that he goes through and hes even mentioned in the description it was 390 spell damage here https://www.youtube.com/watch?v=ay7RU0N-44Y

Currently on elysium I am hitting a pathetic 260-270 judgement of command with 390 spell damage and thesame weapon and same gear. (no sanctity aura)

after doing 30 judgement of commands in 0 spell damage gear I am doing 170-180 judgements. (same as tooptip)

With 204 spell damage hero charm trinket + 390 spell damage gear (594)  I am hitting 305-310 judgement of command. He is hitting far higher judgements than us on this server without using his trinket. Fix please

The coefficient needs to be higher since you can see him hitting higher than elysium. 
I am currently dissapointing with thinking we are being cheated of retail damage. The damage is also coming from late vanilla. There are many changes, for example

I am only gaining 90 damage on judgement of command right now with 390 spell damage

which is a 23% coefficient with lothamir. 
judgement of command should scale at 43%. 

http://i.imgur.com/CGLtyjo.png
This paladin is doing 244 judgement of commands at 58 spell damage in 1.12
watch his video please

Reply

Sorry Imba but I dont agree. Most of the dates are late with dernos too and As we can see this paladin in late vanilla (naxx patch) and is in exactly 163 spell damage gear, and 313 with the flask meaning he should hit
non stunned judgement of command for 305-340 (he states before crusader strike in description)

and 335-374 with sanctity

http://db.vanillagaming.org/?spell=20920

is hitting a 449 judgement of command hit 2:43 and up to 500 judgements (vengeance added)

he hits commands many times in the mid 400s

nightfall procs at one moment later and the damage is higher.

his crits are very very high and this means it doesnt scale at a measly 23% change this please

https://www.youtube.com/watch?v=VEs9coe5XRg&spfreload=10

[...]

PR #2152, opened 2017-08-18
Fix to Judgement of Command scaling.

GuybrushGit:
Fixes https://github.com/elysium-project/server/issues/1136


Mostprovidedlinks:
Thepparker please can you fix JoR too. Its not just JoC but also JoR.

Jor is scaling at 43% and not 50%

PR #2154, opened 2017-08-19
Implement battleground queue randomization.

brotalnia:
Adds 2 new config options that can be used to make organizing premades harder.

- BattleGround.RandomizeQueues
If set to 1 the positions in the queue get randomized, so that the order in which players queued does not matter. When there are enough players to start a battleground, it will invite random ones, not the first to queue.

- BattleGround.GroupQueueLimit
Limits how many people can queue as a group. If the group size is bigger than the limit, the players get queued individually, instead of as a single group.


tribe1234:
This could be the worst idea anyone has ever had.

It will destroy pvp on elysium.

cocaine944:
That's because alliance got 3% win rate? Hail hydra.

bug1991:
This has to be single handedly the most retarded change ever proposed.

Brunheda:
So instead of increasing the likelihood that premades are more likely or even only allowed to face one another with open spots getting filled in with pugs, you have made premading impossible. Incredible.

PR #2157, opened 2017-08-19
Fix for compile error, options for searching by account as well as name and guid for listing deleted characters command

gearselysium:
0 is not valid for std::function<uint32()> , replaced with a closure that returns 0 instead.


brotalnia:
What compiler are you using? I am not getting any compile errors in that file. Also i don't think that the "IsMapServer" option was intended to control whether maps should be loaded or not. This option is under the Clustering section and it is turned off by default.

gearselysium:
VS 2015. If you don't have maps, mangosd won't run because it fails checking for those maps on startup. This is with the mangosd.conf setting mmap.enabled = 0

gearselysium:
I changed the config flag its checking to CONFIG_BOOL_MMAP_ENABLED

brotalnia:
Now you are confusing maps with mmaps, those are different things. Maps are mandatory, mmaps are not and the server will run without them. You can tell by the function name "ExistMapAndVMap" that it is not checking mmaps. You are not supposed to run the server without maps.

gearselysium:
The point is that vmaps are not optional, despite being able to turn off the vmmap features, and the game runs fine without them, that check at start up prevents the server from spinning up. There needs to be a way to disable it for development that does not have vmmaps extracted or for installations which are not using vmmaps.

gearselysium:
Using the MMAP flag for this works and seems harmless since it is only disabling the startup zones check for the races and not interfering with anything else. It is this check that stops the server from spinning up otherwise.

brotalnia:
Then go ahead and extract the vmaps? Don't repurpose the mmap config option for something it's not supposed to do. It works is not a good reason, you can make every other config option do the same thing too, that doesn't mean it makes sense to do it. Mmap options are for mmaps, not vmaps. During the time you spent editing the core to bypass the maps and vmaps check, you could have extracted them 10 times already. You are not meant to run the server without them, this is not a sandbox. You will have things being broken when they won't be if you run it with everything extracted properly. Allowing to bypass those checks will only lead to people reporting non-existing bugs. For example i've had people report to me that Blink is broken and does not teleport you, but after asking them a bunch of questions, it turned out they ran the server without mmaps, and everything was fine once they extracted those.

https://github.com/elysium-project/server/files/852498/Elysium.Extractors.18.March.zip

martinjlowm:
@brotalnia you should quote yourself to some README :P

gearselysium:
You're still missing the point. Whether vmaps are extracted or not is irrelevant. If the core is setup not to be using vmaps, it shouldn't be trying to load them - that is a bug. I don't know why this check was added in this way to begin with - ExistMapandVMap appears to be only called from that one place - it should have been correctly flagged to start with, instead of just making the assumption that vmaps will always be present.

PR #2161, opened 2017-08-19
Mekgineer Thermaplugg bomb spawn limit.

GuybrushGit:
Limited to 5.


src/scripts/eastern_kingdoms/dun_morogh/gnomeregan/boss_thermaplugg.cpp
```diff
@@ -38,6 +38,7 @@ enum
     SPELL_WALKING_BOMB_EFFECT           = 11504,
 
     NPC_WALKING_BOMB                    = 7915,
+    MAX_WALKING_BOMB                    = 5,
```
thepparker: Should be 6, aka 1 bomb from each face (could also just directly check MAX_GNOME_FACES)

GuybrushGit: Done

Issue #1757, opened 2017-07-27
strange interactions between Will of the Forsaken (undead racial) and mind control (priest spell)


a-big-dumb-baby:
https://my.mixtape.moe/qljbog.jpg

As you can see, if precasted before the hostile priest finishes their cast, reports "immune", and the effect doesn't land.  This is obviously WAD.  But weird things start to happen when you try to cast it while already under a mind control effect (not limited only to priest mind control spell, but it was the easiest to test.)

https://my.mixtape.moe/iqkfws.jpg
https://my.mixtape.moe/ldwgfa.jpg
https://my.mixtape.moe/dwuxjm.jpg

Using it while already under a MC effect is a crapshoot.  Sometimes it won't cast at all, and when the effect breaks on its own, you are unable to cast anything until you press esc to stop attempting to cast will of the forsaken.  You can also see in this screenshot that using WoTF while under an MC effect sometimes does work but bugs the player out in other ways.  In this instance neither of us were able to attack each other even though the duel was still going.

Ripper007:
I happen to me also during a game of WSG, after i got mind controlled, i couldn't use the racial at all. After the mind control ended i couldn't use any abilities.

Balrook:
Confirmed, happends alot in BG's. When mindcontrolled and trying to push "will of forsaken" nothing happends. After it wears of, one cant do any other spell or abilitys, cause the systems thinks the will of forsaken is about to cast. Pressing "ESC" on keyb cancels the casting of "will of forsaken" and its back to normal.

Issue #2027, opened 2017-08-10
Will of the forsaken does not work when being mind controlled


kv123:
According to wowpedia, mind control is a form of charm effect, thus undead racial will of the forsaken should make player immune to those.

I remember this working correctly at Nostalrius, Anathema, but it doesnt work on Elysium.

Aorynn:
I find that spamming the ability, can cause you to break the mind control, nonetheless it doesn't seem to function 100% as intended.

PR #2174, opened 2017-08-20
Fix to spell casting while mind controlled.

GuybrushGit:
Couldn't find a good way to select all spells castable under mind control, had to hardcode the spell ids.

Fixes https://github.com/elysium-project/server/issues/1757 https://github.com/elysium-project/server/issues/2027


thepparker:
I think the MC caster check itself needs another rework after the last MC changes, it's on my TODO. It should work something akin to the following:

- If you are the MCer, you can only cast your spells
- If you are the one who is MCed, you can only cast your spells

Note that normal pet cast handling is done in HandlePetCastSpellOpcode in PetHandler.cpp

Normally the cast checks in Spell::CheckCast should take care of whether the spell is actually castable, but I suspect these changes were added for anti-dos measures. Basically, any spell which grants immunity to the Charm mechanic should be castable while you are MCed

GuybrushGit:
I tested some and found non-castable charm immunity spells, but they were npc spells so it might be irrelevant.

A better solution would be removing the GetMover and GetCharmer from _HandleCastSpellOpcode_ and have every spell use \_player, see my last commit. Pet spells use a different function so its not needed right? Unless I'm missing some edge case.

thepparker:
Yea, I'm not sure why these checks are there. There is obviously something going on since it's checking for creature casts when these commands should only be from players. There's also issues with casting to Creature when the caster is TYPEID_UNIT (it crashes), so I think it requires some investigation.

PR #2177, opened 2017-08-20
Allow Mac PPC clients to log in

a-chaudhari:
Admittedly, there are few mac ppc users out there these days.  But Mangos seems to support it and Blizzard shipped ppc binaries all the way until Cata.

It's a small change and it works fine on my local server.

### DIFF
--- a/src/realmd/AuthSocket.cpp
+++ b/src/realmd/AuthSocket.cpp
@@ -764,7 +764,7 @@ bool AuthSocket::_HandleLogonProof()
     }
 
     // reject credentials on unexpected build to confuse custom client authors
-    auto const approvedBuild = _platform == X86 && (_os == Win || _os == OSX);
+    auto const approvedBuild = (_platform == X86 || _platform == PPC) && (_os == Win || _os == OSX);
 
     ///- Check if SRP6 results match (password is correct), else send an error
     if (!memcmp(M.AsByteArray().data(), lp.M1, 20) && pinResult && approvedBuild)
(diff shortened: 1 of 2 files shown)


BBCnewstoday:
I think, this should be rolled back! Nobody is using PPC nowadays! And its open more holes for hackers! Since warden needs updates for PPC and it will ignore any cheaters that are using PPC!!!

a-chaudhari:
1. That’s quite the statement to make unless you’ve access to Elysium’s anti-cheat implementation source code

2. From what I see the platform is only checked in realmd login.  the game server doesn’t seem to bother with it.  I’m assuming blizz coded both the PPC and X86 clients to have the same server-client interface for both warden opcodes and normal game traffic.  Wouldn’t make sense otherwise.
  
3. even if blizard’s PPC wow client had some magical bug that rendered all of the Elysium anticheat useless, it’d be trivial to modify/spoof an authentication request in-client or on-wire to show up as X86 and use the PPC client anyway.  It’s literally just a string the client passes up during auth.

BTW, Stock mangos doesn’t even bother checking it.  I’m assuming this check in elysium is a remnant from Nost’s source when their code was closed.  This type of check is useless in an open source project when any script kiddy can look at the code and see what the required values are.

4. Obviously PPC isn’t a high demand platform these days.  But I know I’m not the only one that has tried to play on Elysium with a PPC client.  First result for “elysium wow ppc”: https://forum.elysium-project.org/topic/33411-mac-client-on-powerpc/  I bet a bunch of people have tried over time.

BBCnewstoday:
How Warden works? 
First it checks the platform WoW runs, after this its set detect methods, based on platform.
Many years ago cheaters on windows just masked self as Mac, so Warden thinks they are using Mac, so can't detect Windows cheats, when you mask your Windows for Mac.
Now, cheaters can do same with PPC ;-) Just mask your windows as PPC.

Issue #2032, opened 2017-08-10
Viscidus Blobs Moving WAY too fast again


Forest-runklebc:
https://youtu.be/iQ7bzxOHexY?t=8187 - Current Elysium Version

https://youtu.be/QspWtDufeZ0?t=75 - Retail version

fgsfdss:
Going from the reference video it seems to be missing a visual animation as well when shattering.

Issue #1527, opened 2017-07-17
Viscidus blobs moving too fast.


rockmusicalways:
After Viscidus is shattered and splits into many blobs, the blobs are reforming WAY to fast.  
~20 sec clip showing my guilds brief attempt earlier today (16/7/2017) - sorry for any bad language in voice com's.   https://www.twitch.tv/videos/159712212

StadenElysium:
Fix incoming on Wednesday.

kampeador:
From what i saw in code: it is still not fixed. 
Retail: speed interpolated linearly from 0 to 1.
You can clearly see that in this video:
https://youtu.be/Kc8yIFZjV6s?t=132

Current implementation:
Hacky mess. Their speed starts at 1.0 and increases by 1.5 every second i.e 1.0, 2.5, 4.0, 5.5... which makes no sense at all.
Once i'll find some free time, i'll rewrite this.

rockmusicalways:
What kampeador said is correct. Guild attempted Visc again today with the same results as the original video clip I posted.

QAkvastskaft:
Fixed here: https://forum.elysium-project.org/topic/53637-development-update-08092017/

PR #2178, opened 2017-08-20
Viscidus improvements.

kampeador:
- Corrected Boss spawn position.
- Corrected text emotes.
- Explosion(visual effect) is now visible.
- Rewritten glob behavior. Implemented smooth movement and acceleration.
- Rewritten toxic cloud behavior. Timer is now random(between 30-40 seconds) and boss doesn't cast it directly. It should be used only by triggers.
- Implemented a visual effect which predates a toxic cloud.


GuybrushGit:
https://github.com/elysium-project/server/issues/1527 https://github.com/elysium-project/server/issues/2032

src/scripts/kalimdor/silithus/temple_of_ahnqiraj/boss_viscidus.cpp
```diff
@@ -333,33 +429,31 @@ struct boss_viscidusAI : public ScriptedAI
             else if (m_uiHitCount >= HITCOUNT_SLOW_MORE)
             {
                 if (m_uiHitCount == HITCOUNT_SLOW_MORE)
-		{
+		        {
```
thepparker: There are many tabs in here, can you replace them all?

PR #2264, opened 2017-08-26
Fix to Emperor Dagran split pulling

GuybrushGit:
The Emperor can't be chained to the mobs since they already have their own chain groups.
Added a call for help timer.


thepparker:
How is the initial pull done? I think it should just pulse generically for any mob that should initially pull the surrounding NPCs

GuybrushGit:
It does a CallForHelp on agro. 
I checked and didn't find any other mob that does this. They all use the linked groups system, which is not usable in this case because the mobs in the Emperor room have their own groups among themselves.

PR #2265, opened 2017-08-26
Fix to Execute not usable at 20%

GuybrushGit:
Fixes https://github.com/elysium-project/server/issues/1512


elysium-asura:
Logic behind this change is incorrect.

`ModifyAuraState(AURA_STATE_HEALTHLESS_20_PERCENT, GetHealth() < GetMaxHealth() * 0.20f);`

That is the only issue here. The check only cares currently if it is LESS THAN 20%. It could be altered to:

`ModifyAuraState(AURA_STATE_HEALTHLESS_20_PERCENT, GetHealth() <= GetMaxHealth() * 0.20f);`

That would solve the issue.

Obviously, the rest of your changes are cosmetic and exactly the same, but with < instead of <=. This should have been a dead giveaway to the actual issue.

Please open a new pull request with the correct fix.

PR #2346, opened 2017-09-01
Fix memleak in Unit::CleanupDeletedAura(delete auras in use case)

ryancheung:


thepparker:
A better solution would be finding why auras are still marked as in use at this stage, especially if they are actually being used by something and we delete them...

ryancheung:
Deleting auras in use is inevitable, e.g. A summoned creature must be removed after death immediately, and the death was caused by a DOT.
@thepparker

ryancheung:
Maybe you can set a delay, 100ms for example to remove the summoned creature. But this would avoid deleting auras in use 100%, since world update delay may exceed 100ms.

thepparker:
If the creature dies and needs to be instantly despawned, the auras should be removed (on death, or before despawn). This is more like removing the symptoms rather than the cause.

ryancheung:
There is a case: 
Given a creature summing adds, adds should be despawned immediately in `ScriptedAI::SummonedCreatureJustDied`(or `JustDied` and despawning self).

If you want to remove the aura before creature despawn, current core has no way to do it.

And I believe the way deleting aura before despawn is too complicate, it will need to change the creature despawning mechanism.

thepparker:
When creatures are unsummoned (or force despawned), Map::Remove is called, which calls Unit::CleanupsBeforeDeleted, deleting existing auras and cleaning any expired ones up. If an aura is still in use at this stage, then that needs to be changed

Issue #1769, opened 2017-07-27
Quest Field Duty


Fnt:
After talking with the quest mob, a script is launched, but the quest is not executed.

Mostprovidedlinks:
Fix this please

QAkvastskaft:
Fixed here: https://forum.elysium-project.org/topic/55357-anathema-updates-27092017/

Issue #902, opened 2017-05-28
Field Duty (CC Rep)


Ravyd:
Whatever happened to the Field Duty quests awhile ago completely destroyed the quest chain. Hundreds of people were a few badges away when Krug got completely glitched out and now requires you to kill the Hunter-Killer for every task turn in.

I finally got a group today to kill the mob and thought it would be taken care of. I made sure that I started the quest and that we weren't in a raid group. We killed him (I also didn't die) and I turned it in. I come back with another Field Duty and once again he wants me to kill it again to turn in.

Hunter Killer should be killed only once to enable task assignments.

You cannot get new assignments, killing that elite every time is completely unreasonable.

Ravyd:
Anathema realm. Possible hotfixes: buffing krug to solo Hunter-Killer, making Hunterkiller level 1 non elite with instant respawn next to quest giver

Saltynow:
What's the matter in putting it back to what it was before? You used to start the quest chain in Cenarion Hold @Windcaller Kaldon, ran to Krug and got your field papers signed . There was some issue, in which you had 2x Prepared Field Papers in your bag after talking to him once and you had to delete 1 to turn it finally in at Krug. Guess they wanted to fix this and sth went wrong there. Now you have the problem Ravyd just explained 19 days ago: the mob spawning (if at all) is way to strong to kill it solo, which wasn't supposed to happen over and over again anyways. This way ppl can't turn in their last field duty badges. Help out on that guys!

Aleksxz:
Video from bug tracker : https://www.youtube.com/watch?v=pOF3rGqFAwA

Sometimes the event doesn't even launch.

QAkvastskaft:
Fixed here: https://forum.elysium-project.org/topic/55357-anathema-updates-27092017/

Issue #854, opened 2017-05-25
Krug Skullsplit (npc=15612) does no longer give signed field duty since the event is bugged.


SiguDS:
Waited a long time at Krug Skullsplit to see if an event would happen or something but he seems just not to be giving out signed field duty papers. I have not been able to get it done since the wednesday update. Retaking the quest etc does not seem to change this.

SiguDS:
Seems the event just tries to start but nothing comes from the hive and he just kinda stops in his tracks.
http://i.imgur.com/qmXT5nb.jpg

SiguDS:
After the event then "reset" and stuff walked back you still could not get the field duty quest. http://i.imgur.com/qmXT5nb.jpg

SiguDS:
Seems that the mob sometimes comes from the hive but when it kills krull the next time it doesn't show up.

SiguDS:
Still kinda buggy it either evades or it despawns. Only 1 out of 12 events i have been able to get it done. Video to show: https://www.youtube.com/watch?v=nLWvfu7dey4&feature=youtu.be

QAkvastskaft:
Fixed here: https://forum.elysium-project.org/topic/55357-anathema-updates-27092017/

PR #2348, opened 2017-09-01
Fixes to Horde Field Duty event.

GuybrushGit:
Related Issues https://github.com/elysium-project/server/issues/1769 https://github.com/elysium-project/server/issues/902 https://github.com/elysium-project/server/issues/854

- Set Hunter-Killer as an active object to prevent the mob from not spawning when no one is in his grid.
- Increased his speed and despawn time so he doesn't despawn before the event starts.
- Decreased Skullsplit respawn time ( "he respawns immediately after the fight" http://web.archive.org/web/20071021051539/http://thottbot.com:80/c15620 ).
- Removed all the debug messages that were still being sent.
- Increased the time between events from 30 secs to 30 minutes. Seems ridiculous that you have to kill the 350k hp mob every 30 seconds, its probably a left over from whoever was debugging it.

I didn't modify Hunter-Killer's stats but I noticed he's using the stats from vanilla WoWWiki, while the Alliance boss is using the stats from the TBC Bestiary, which has both mobs with far less HP and damage (WoWWiki has no stats for the Alliance boss).
Should consider using the TBC stats for the Horde boss as well, so both factions are equal.


thepparker:
Let's go with the HP values of the Horde NPC for the Alliance mob as well, up it to ~350k and add a bit of damage if necessary

GuybrushGit:
Ok equalized the stats for the Alliance boss and fixed the Horde boss Cleave spell.

Issue #2410, opened 2017-09-05
Loot range should be increased


BrynjaDarrow:
Atm the range in which you have to be to a boss to be able to loot him during raids is too small. This can make it impossible for people to loot the boss after he is downed, just because they were out of range.

Example:
If your mages move away during the mage classcall for example and Nef then dies, he may be impossible to loot for them. Doesn't matter if you are dead or alive btw. 

I reckon the range, in which you are eligible to loot should at least be as large as the largest bossrooms.

malekoydaerb:
Seconded

Last night Intoxicated (anathema alliance) killed Ouro. Instance ID 405. Was a messy kill but a kill none-the-less with 25 deaths. Roughly 15 of those 25 were unable to receive loot form Ouro, despite them being in his room, because of range from him when he died. I was killed by a sweep right before he did his last submerge and when he came back up, was on the other side of his room. Because of that range, I was unable to even view loot on him, let alone receive anything via masterloot.

I have never in all my years playing wow, even on other buggier private servers, encountered this issue. As well, it's not just Ouro, I have personally seen it occur on bosses in UBRS(Drakkisath), MC(Garr and Rag), Ony, and BWL(Nef). 

Players should not be penalized, and total lack of loot is a huge penalty, for simply having the bad luck to die within the bosses room, but further than 40 yards from the boss.

Cabro:
To be honest, if you were active in a boss kill in a raid, then it shouldn't matter where in the whole instance you are, when the loot popups you should get the popups also. Same for Master Looter, you should be in the assign menu.

malekoydaerb:
@Cabro if you release, you are not in the instance, and thus dont get loot. that is a known thing and not an issue, your fault for releasing. the issue we're describing is coming about when you are present for the boss, helping kill it but dying during the fight before said boss, and being far away from the boss when he dies and thus not getting any loot. 

my direct experience last night had myself and about 10-15 others unable to see any loot on the boss. as well, the loot master was unable to send anything to us and was getting an invalid target error when they tried.

yankopetrov:
We had the same problem with Nef. Couple of raiders died in phase 1 and they was unable to loot later.

Cabro:
[...]

PR #2433, opened 2017-09-06
Removed distance limits from raid loot.

GuybrushGit:
https://github.com/elysium-project/server/issues/2410 
Changed raid loot to go out for everyone in the instance and increased loot range on Gandling and world bosses.


thepparker:
I think this was limited for a particular reason, why I'm not sure since there's no commenting, but it was not originally limited so severely here. Allowing loot from any range inside a raid instance would open up issues with selling items and having people just sit at the front of the instance chilling for their gear, which I'm not sure is something we want to encourage. Checking for the obj in the unit's m_attackers list might be better, as long as healing and other related effects add them to it...

GuybrushGit:
>Allowing loot from any range inside a raid instance would open up issues with selling items and having people just sit at the front of the instance chilling for their gear,

Is that an actual issue? On regular dungeons I get it but on raids I don't see the problem in having people chill at the entrance, they're still taking up a raid spot and low levels can't enter the raid. People now can just chill and put someone on follow its essentially the same thing.

Checking attackers list is too unreliable people can easily drop out of the attackers list.
We can create a boss specific CONFIG_FLOAT_GROUP_XP_DISTANCE but I still think raid wide loot is the best solution.

Cabro:
The problem is that you get the loot window popups even if you didn't participate in the fights. It happened to me in a 5man dungeon, maybe this is happening in raids also, and if it does, it is a bug.
You should only get loot popups from the mobs you actively participated to defeat. A boss is still a mob and the same rule should apply.

PR #2437, opened 2017-09-06
Fix to no loot when master looter not found.

GuybrushGit:
Issue happens when the loot master is offline or in loading screen.
When a master is not found loot type now switches to group loot.


brotalnia:
Someone with a retail sub ought to check if that is how it is supposed to work.

GuybrushGit:
The alternative would be adding the offline master to the allowed looters list, but that would open up the possibility for people to intentionally go offline and make themselves eligible to raid loot, while they're in Ironforge or something.

brotalnia:
What do you think about setting someone else as the master looter when the current one is offline?

GuybrushGit:
Give master loot to a random? Doesn't seem like a good idea.

Cabro:
I remember that on retail you could change Master Looter to another player. That is if ML != RL .
In the case where ML = RL then it would give RL+ML to an assistant if possible, and if no assistants were in raid it would go to a random. Yes, sounds like a bad idea, but it is what it was blizzlike.

gearselysium:
On retail, IIRC, the RL could change the master looter at any time, even if the current master looter was offline. If ML = RL, you were SOL. This would have changed when they added individual loot, which they then mailed to the player if it wasn't picked up.

PR #2469, opened 2017-09-08
Fixes many NPCs that are incorrectly ranked as Rare instead of Rare-Elite

Jakobud:
Many NPCs are incorrectly ranked as Rare when they should be Rare Elites. This fix updates the db tables to fix this issue for all of them.

I examined all NPCs ranked as Rare (rank 4) in the database and cross-referenced it will their info on db.vanillagaming.org and updated all the NPCs that were incorrectly ranked.
Also if you need further proof beyond db.vanillagaming.org, simply fight any of these NPCs in-game. You will find them much more powerful and with much higher HP compared to all normal NPCs near them, obviously indicating that they are a Rare-Elite instead of a Rare.


brotalnia:
Vanillagaming is not a valid source! That is just another private server's database. Only thing of value there is the wowhead comments. You should be checking old versions Allakhazam and Thottbot instead.

Just one example:
https://web.archive.org/web/20060114035722/http://wow.allakhazam.com:80/db/mob.html?wmob=14222
https://web.archive.org/web/20070704184514/http://www.thottbot.com:80/c14222
https://youtu.be/zmeq0dmcq-s?t=6m7s

As you can see Araga is not elite in both databases. Not even in the retail video from 2009. I suspect most of the others you "fixed" are not supposed to be elite either... Go through them again and only update the ones that are elite in Allakhazam and Thottbot versions or retail videos from before Cataclysm (uploaded 2004-2010).

Jakobud:
Thanks, I will go back and re-evaluate this. Need to re-write my scraper to use those old archive pages instead. Thanks

PR #2476, opened 2017-09-09
Extend Deleted Character List Commands

gearselysium:
Extend the deleted character list command to take an account or name switch and allow searching using partial account name match or account id match.


brotalnia:
This commit causes x86 builds on Windows to crash as soon as you try to use any command.

PR #2483, opened 2017-09-09
Fixed Triage quest PVP flag bug

Adamjp29:
There is currently an issue that allows players to attack unflagged players of opposing faction and allows riding mount underwater if a player abandons or fails this quest.


src/scripts/world/npcs_special.cpp
```diff
@@ -492,6 +492,7 @@ void npc_doctorAI::PatientDied(Location* Point)
     }
     else
         // If no player or player abandon quest in progress
+        pPlayer->SetFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_PVP_ATTACKABLE);
```
thepparker: you need to add braces around this statement and Reset() so they are executed in the same block

PR #2492, opened 2017-09-10
Prevent pet pulling Magistrate and techies

GuybrushGit:
Prevents pulling them through gates.


thepparker:
TODO: Implement proper detection for gobjs and prevent pathing through them

Issue #2149, opened 2017-08-18
Darkmoon card: Maelstrom proccing/breaking off CC (e.g. Gouge)


Zonati:
First of all, this is a difficult bug to find evidence for since there seems to be little direct information about how this particular item interacts with Incapacitating effects such as gouge. However, I have done numerous research into other proc items, which work in similar, or in the exact same way, as Darkmoon card: Maelstrom. One such proc is the Storm-shroud 2 set bonus: "5% chance of dealing 15-26 nature damage on a successful melee attack" http://db.vanillagaming.org/?spell=18979 (I'm aware this website isn't considered a valid source, but these are how the procs work on Elysium aswell) This works the exact same way as Maelstrom works, apart from proc rate and damage, http://db.vanillagaming.org/?spell=23686. The difference is that the Stormshroud set doesn't break incapacitate effects on proc (which is blizzlike: Patch 1.11.0 Stormshroud Set: The Lightning effect from this set will no longer break your Gouge; (source: http://shadowpanther.net/roguechanges.htm) However this trinket does. Additionally, comments from February 2006 say that this trinket should NOT proc on gouge: "this procs on melee strikes including instant attacks.  Not positive about a proc on sap, however it will proc on a gouge but does not break the gouge. (source: http://web.archive.org/web/20070515083410/http://www.thottbot.com/i19289 ). This comment was written when WoW was at patch 1.9, however; there are no patch notes atall regarding this trinket, apart from it's release in 1.6, so I assume that it was like that on release aswell.

GuybrushGit:
Stormshroud is breaking CC as well, everything is.

Harkonnen13:
Sad, bud this issue still works on Elysium. The Lifesteal enchant proc breaks Gouge (not really often, but still), as well as the Darkmoon card: Maelstrom.

PR #2506, opened 2017-09-10
Prevent proc spells from breaking CC

GuybrushGit:
Fixes https://github.com/elysium-project/server/issues/2149

> The fire damage from the Blazefury Medallion and the Gutgore Ripper's effect should no longer immediately break Gouge.
> Stormshroud Set: The Lightning effect from this set will no longer break your Gouge.
> http://shadowpanther.net/roguechanges.htm

> Maelstrom will proc on a gouge but does not break the gouge.
> http://web.archive.org/web/20070515083410/http://www.thottbot.com/i19289

Seems like the default behaviour for these triggered proc spells is not to break the CC that triggered them.
Made it so spells caused by SPELL_AURA_PROC_TRIGGER_SPELL dont remove auras with SPELL_ATTR_EX_UNK18 (stun, polymorph, daze, sleep)


thepparker:
I disagree. It seems pretty explicit to me that it should only be spells triggered by casting the CC that should not interrupt the CC immediately.

GuybrushGit:
Yeah ideally you would want to check that, but it would mean that spells would have to store the spell that triggered the spell that triggered them. I can implement it that way if you think its fine.

thepparker:
I think it's worth tracking the spell it was procced from, yea

GuybrushGit:
Fixed it.

Harkonnen13:
Sad, bud this issue still works on Elysium. The Lifesteal enchant proc breaks Gouge (not really often, but still), as well as the Darkmoon card: Maelstrom.

ryancheung:
Is this fix live now or not?

PR #2510, opened 2017-09-10
Fix to aura duration after a long time offline

GuybrushGit:
Fixes the common resurrection sickness issue.


Mostprovidedlinks:
Good job

Issue #2248, opened 2017-08-25
Argent dawn medics curing the player


Sabre142:
They're meant to remove the debuffs you get in Eastern Plaguelands.

http://wowwiki.wikia.com/wiki/Argent_Medic?oldid=374523

PR #2576, opened 2017-09-14
Implemented an Argent Medic behavior.

kampeador:
- Implemented an Argent Medic behavior. https://github.com/elysium-project/server/issues/2248
- Implemented a Diseased Flayer AI.
- A script for Gibbering Ghoul, Diseased Flayer and Cannibal Ghoul (Defenders of Darrowshire quest id:5211 related) has been moved to DB. Original script was implemented wrong. It doesn't matter who kills a ghoul. https://github.com/elysium-project/server/issues/2791


src/scripts/eastern_kingdoms/eastern_plaguelands/eastern_plaguelands.cpp
```diff
@@ -2110,16 +2084,104 @@ bool EffectDummyGameObj_go_mark_of_detonation(Unit* pCaster, uint32 uiSpellId, S
...
+    void UpdateAI(const uint32 uiDiff)
+    {
+        // no point checking for nearby players if npc is in combat
+        if (!m_creature->isInCombat())
+        {
+            // scan for players
+            if (m_cureDeseaseCooldown < uiDiff)
+            {
+                // search area for nearby player characters
+                Map::PlayerList const& players = m_creature->GetMap()->GetPlayers();
```
thepparker: You should use a grid check here, maybe within the range of cure, to find players rather than checking the entire map

kampeador:
You're definely right for creatures, but players are stored separately inside a linked list for each map. I've implemented both versions.

UpdateAI() is called only if map cell is active i.e 1+ players are inside it or some objects are in active state(not in this case).

Timers are for 1 NPC. There are only 2 of them in game, so they will be doubled.
Players are checked every 5 seconds.

Linux gcc 7.1 -O3
Lower is better:

Map:
1 player near:    0.000875ms(dist < 16)
1 player far:       0.000252ms(dist > 16 inside cell)
20 players near: 0.002248ms(dist < 16)
45 players near: 0.004945ms(dist < 16)
45 players far:    0.002269ms(on the map or dist > 16)

MapCell:
1 player: 0.001551ms
20 players: 0.004276ms
45 players: 0.007756ms

Map checks are faster for low player count and/or small maps, but for this case i'll probably optimize my MapCell version.

thepparker:
Keep in mind that the 'map' is not just EPL, it can be the entire continent, or just the top portion of EK if continents are instanced. You could have a thousand players or more in it. Undercity will always also be in the same map which guarantees some population. There are nearby player searchers that you can easily utilize.

kampeador:
**thepparker** Thanks for your review and suggestion. My new version is much faster.

kampeador:
I'll update this with more blizzlike behavior.

kampeador:
I've checked this NPC on retail. It was quite challenging to trigger this behavior, because disease fade faster than i reach that medic...
I've done some "overengineering" here. There are other NPC/creatures which have similar behavior and i'll reuse this logic for them.
Argent Medic has a simple gossip option.
I'll update this soon.

brotalnia:
Here is the localization for the gossip option from trinity db.

INSERT INTO `locales_gossip_menu_option` (`menu_id`, `id`, `option_text_loc1`, `option_text_loc2`, `option_text_loc3`, `option_text_loc4`, `option_text_loc5`, `option_text_loc6`, `option_text_loc7`, `option_text_loc8`, `box_text_loc1`, `box_text_loc2`, `box_text_loc3`, `box_text_loc4`, `box_text_loc5`, `box_text_loc6`, `box_text_loc7`, `box_text_loc8`) VALUES (7175, 0, '질병에 걸렸습니다. 치료를 부탁합니다.', 'Je suis malade. Soignez-moi, infirmier.', 'Ich bin krank. Sanitäter, bitte heilt mich.', '我的伤口感染了。请治疗我，医生。', '我生病了。請救救我，醫生。', 'Estoy enfermo. Cúrame, por favor.', 'Estoy enfermo. Cúrame, por favor.', 'Мне очень плохо. Пожалуйста, вылечите меня.', NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL);

kampeador:
Thanks. I've added it.

Issue #2588, opened 2017-09-15
[Elysium] - Message to the Wildhammer missing.


lovemonkeyz:
https://github.com/elysium-project/server/issues/1443 was closed saying the quests can be picked up. Message to the Wildhammer is still missing, Otho Moji'ko offers no quests. WDB has been cleared, no change. I checked while in Hinterlands and no one else was able to get it, so this is an ongoing issue.

PR #2602, opened 2017-09-16
Fix for Otho Moji'kos Missing Quests for bug#2588

watapus:
Not sure if I submitted this correctly but fix for Bug #2588 

-Quest: Separation_Anxiety Was not able to replicate this issue, quest existed when I pulled the latest Database and Souce

-Quest: Message_To_the_WildHammer - Changed NPC Flag to 6 from 4(Vendor and Quest Giver), also updated the Type of quest as having it set to PvP was causing it not to show. Also updated the Objectives to the correct values. Previously every Objective was 15 kills when it should be 8/5/5/2.

Source: http://wowwiki.wikia.com/wiki/Quest:Message_to_the_Wildhammer

### DIFF
--- a/sql/migrations/20170916172553_world.sql
+++ b/sql/migrations/20170916172553_world.sql
@@ -1,5 +1,7 @@
+INSERT INTO `migrations` VALUES ('20170916172553');
+
 UPDATE quest_template
-SET ReqCreatureOrGoCount1 = 8, ReqCreatureOrGoCount2 = 5, ReqCreatureOrGoCount3 = 5, ReqCreatureOrGoCount4 = 2, Type = 0
+SET Type = 0
 WHERE entry = 7841;
 
 UPDATE creature_template
--- a/sql/migrations/20170916172553_world.sql
+++ b/sql/migrations/20170916172553_world.sql
@@ -1,0 +1,9 @@
+UPDATE quest_template
+SET ReqCreatureOrGoCount1 = 8, ReqCreatureOrGoCount2 = 5, ReqCreatureOrGoCount3 = 5, ReqCreatureOrGoCount4 = 2, Type = 0
+WHERE entry = 7841;
+
+UPDATE creature_template
+SET npcflag = 6
+WHERE ENTRY = 14738
+
+


GuybrushGit:
The 15 kill requirements were right, wowwiki data is for retail not vanilla:
https://web.archive.org/web/20070406221444/http://www.thottbot.com:80/q7841

watapus:
I see will be sure to check better, removed the Kill Objective Changes.

sql/migrations/20170916172553_world.sql
```diff
@@ -0,0 +1,11 @@
+INSERT INTO `migrations` VALUES ('20170916172553');
+
+UPDATE quest_template
+SET Type = 0
+WHERE entry = 7841;
```
brotalnia: This quest was marked as PvP on retail. It's not a bug.
https://web.archive.org/web/20060525194545/http://wow.allakhazam.com:80/db/quest.html?wquest=7841
https://web.archive.org/web/20070406221444/http://www.thottbot.com:80/q7841

The reason it cannot be taken is this check here: https://github.com/elysium-project/server/blob/development/src/game/Objects/Player.cpp#L12210

It should probably be removed, as it does not make sense. It checks if the quest is a pvp quest, if it has a quest level, and if the quest level is below the player's level. If all those conditions are true, it makes it so you can't take the quest. There is no such check in cmangos and mangoszero.

sql/migrations/20170916172553_world.sql
```diff
@@ -0,0 +1,11 @@
...
+
+UPDATE creature_template
+SET npcflag = 6
+WHERE ENTRY = 14738
```
brotalnia: This npc (Otho Moji'ko) already has `npcflag` set to 7. That includes flags 1, 2 and 4. So what your query actually does is remove his gossip flag.

brotalnia: Also you should be careful with the capitalization of column names. It is "entry" not "ENTRY". That will produce and error on some versions of mysql.

GuybrushGit: That pvp check is there to prevent players from taking the bg turn in quests from lower brackets.

sql/migrations/20170916172553_world.sql
GuybrushGit: That pvp check is there to prevent players from taking the bg turn in quests from lower brackets.

brotalnia: I thought that may be the case, but there has to be another way of implementing that restriction then. The quest is level 48 in every db, so obviously that's not how blizzard prevented players from taking the lower bracket quests, or this one wouldn't have been available past 48 on retail too. We should figure out what was the correct logic for this.

thepparker:
Closed per the comments, this is not the correct way to rectify either of the reported issues.

Issue #2503, opened 2017-09-10
(Thorium) Shield Spike incorrect damage


fidbugs:
Thorium shield spike damage is incorrect. It seems that it is affected by armor reduction and it is not randomized as tool tip suggests. That being said i suppose it is same with all shield spikes(mithril, iron, etc.).

Every time you reequip shield with attached shield spike damage changes and remains constant until next removal and equipping. For instance - you equip shield, it will roll random number and you will be dealing that amount of damage until you reequip shield, and after that it again gives you certain amount of damage you will do on block.  Damage is not randomized/rolled between 20-30 on every block you make.

Second part of issue is that damage done is affected by armor of target or by some sort of reduction mechanism - as i am unable to get damage(with current mechanic) between 20-30, from what i have tested it goes from 11-21 - which shows some kind of reduction.

There is no mention of such reduction involved with spikes damage on thottbot or allakhazam. It clearly states multiple times that damage is 20-30. Also video material shows no such reduction.

Recepie: Thorium Shield Spike

http://web.archive.org/web/20071114220407/http://thottbot.com:80/i12692

http://web.archive.org/web/20080116012439/http://wow.allakhazam.com:80/db/item.html?witem=12692&locale=enUS;source=live

Thorium Shield Spike

http://web.archive.org/web/20070504065243/http://wow.allakhazam.com:80/db/item.html?witem=12645&locale=enUS;source=live

http://web.archive.org/web/20071114220402/http://thottbot.com:80/i12645

I suppose randomization of damage on hit is obvious and self-explanatory based on tool tip description.

Video evidence tho from TBC, it confirms correct mechanics since no change to item(or whole family of items, spikes) has been made made.

https://www.youtube.com/watch?v=lJVVU1En3Bg

Mostprovidedlinks:
Agree to this

PR #2622, opened 2017-09-17
Shield spike damage

GuybrushGit:
Fixes https://github.com/elysium-project/server/issues/2503


thepparker:
This prevents auras from scaling with attack power/spell power. It probably should be re-rolled on hit, but needs to scale. In the case of spikes being mitigated by armour, perhaps we need to add specific handling for them (i.e. give them the custom attr SPELL_CUSTOM_IGNORE_ARMOR)

GuybrushGit:
Check my last commit I believe its all good now.

Added just the spell damage scaling, everything else is unecessary:
There are no Aura 43 spells with Melee / Range dmg class so we don't have to do any AP scaling.
The only non-magic spells are the spikes so we don't have to do armor reductions.
Currently these shield spells are critting but they shouldnt, we can leave the crit calcs out as well: [source](http://wowwiki.wikia.com/wiki/Lightning_Shield?oldid=360145).

fidbugs:
While you are at it, same thing is happening with Crest of Retribution.

http://db.vanillagaming.org/?item=13375#wh-comments

http://web.archive.org/web/20071013060112/http://thottbot.com:80/i13375

I have an issue finding any retail evidence, but it behaves currently same as spikes - its mitigated by damage and not randomized. Its effect should follow same rules as every other spike, since it is basically bare shield with spike effect.

Person calculating possible reflection damage takes Crest of Retribution in same equation with Retribution Aura and spikes, thus same behavior id assume.

Issue #2379, opened 2017-09-03
Blessing of sacrifice is not working on crystal of zin malors DoT


Itembugs:
Paladin is being hit for DoT damage by crystal of zin malor.

Blessing of sacrifice is meant to work when someone 

takes auto attack (white) melee damage
takes special (yellow) melee damage
takes auto attack (white) ranged damage (with a bow, etc)
takes special (yellow) ranged damage (with a bow, etc)
recieves beneficial spell
receives a harmful spell
takes periodic (DoT) damage

1. It works on player ticks/dots

"transfering 55 damage taken per hit to the caster."

Corruption and this item have thesame aura

http://db.vanillagaming.org/?spell=25311

it also says receives a harmful spell
takes periodic (DoT) damage

meaning it can be any spell from anything.

Itembugs:
A flag carrier warrior and a healbot paladin is the reason this is important. Fix please.

Itembugs:
There is a reason for this, this is for warrior and paladin flag carrying

Itembugs:
Please @guybrushgit  can yiu add to your fixes ty

Itembugs:
Blessing of sacrifice is meant to work when someone

takes auto attack (white) melee damage
takes special (yellow) melee damage
takes auto attack (white) ranged damage (with a bow, etc)
takes special (yellow) ranged damage (with a bow, etc)
receives a harmful spell 
Recieves a beneficial spell
takes periodic (DoT) damage

Itembugs:
Blessing of sacrifice is meant to work when someone

takes auto attack (white) melee damage
takes special (yellow) melee damage
takes auto attack (white) ranged damage (with a bow, etc)
takes special (yellow) ranged damage (with a bow, etc)
receives a harmful spell
Recieves a beneficial spell
takes periodic (DoT) damage

I have a site showing this

PR #2635, opened 2017-09-18
Fixed self damage not being transfered with Sacrifice / Soul Link

GuybrushGit:
Related to https://github.com/elysium-project/server/issues/2379

There was a "only split damage if not damaging yourself" rule in the code but I found no reason behind it or sources that mention it.

There are reports of both Sacrifice (SPELL_AURA_SPLIT_DAMAGE_FLAT) and Soul Link (SPELL_AURA_SPLIT_DAMAGE_PCT) working with Hellfire:
http://www.ownedcore.com/forums/world-of-warcraft/world-of-warcraft-exploits/211324-halaa-token-bug-mounts-4.html#post1391352
http://www.ownedcore.com/forums/world-of-warcraft/world-of-warcraft-exploits/174688-warlock-arena-get-infight-immun-against-sap.html

I removed the self damage rule but we can also keep it and add an exception for Hellfire.


Itembugs:
blessing of sacrifice is still not working with crystal of zin malor

Issue #2597, opened 2017-09-16
Mind Control/Dominate Mind effects should be dispellable (Need Info)


ohganot:
Lucifron adds cast a spell called Dominate Mind which should be dispellable.

Multiple guides from 2005 state that it should be dispelled as fast as possible:
http://www.asga.co.uk/forum/index.php?topic=77.0
http://www.tentonhammer.com/guides/mc-boss-guide-boss-1-lucifron

This might also apply to Mind Controls in general, but need more research done for that.
Patch 1.4.0 notes state that:
"Charm and Polymorph mechanics can now be reactively countered by spells and abilities (e.g. Druid Shapeshifting will now cancel Polymorph effects)."

GuybrushGit:
So when using dispel on an MCed friend, it can dispell the MC debuff in adition to dispelling a random buff?

Does the MC debuff have dispell priority? Do you have to spam dispell and remove random buffs from him until you get lucky and remove his MC?

Cabro:
If i remember something vaguely about this, is that paladins couldn't dispel MC because it was a hostile spell, while priests and shamans could.
This mechanic of dispelling while MC effects is a special one and should be dealt thoroughly.

ohganot:
Yeah, dispells worked differently on MCd targets. From what I understand, dispelling a mind controlled target would only dispel the mind control, but only offensive dispells would work (Purge, Dispel Magic, Devour Magic).
The video is quite clear for the Lucifron add Dominate Mind spell, but I'm not 100% sure how it should work for priest Mind Control.

Cabro:
As far as i remember, it all depended on the faction of your MC target. As in, if your target was same faction as you and was MC, it would dispel the MC if you were priest, shaman or warlock pet. If you were paladin, nothing dispelled.
If it was opposite faction, it would dispel his beneficial buffs (stamina, intellect etc). Again, not if you were a paladin.

Fiskehatt:
## Some extra evidence:

Lucifron - Molten Core
http://www.ownedcore.com/forums/world-of-warcraft/world-of-warcraft-guides/1004-molten-core-dummies.html (2006-06-10)

> First, the guards randomly Mind Control people. Priests need to dispel this ASAP, but because MC'ed players tend to use their long-duration abilities, WARRIORS ABSOLUTELY MUST BLOW THEIR INTIMIDATING SHOUT BEFORE THE BATTLE. If they do not, they risk getting controlled, running into the middle of the raid, and fearing everyone.

[...]

PR #2682, opened 2017-09-21
Fix to Mind Control dispels

GuybrushGit:
Issue https://github.com/elysium-project/server/issues/2597
Also fixes dispelling the MC caster not interrupting the MC.

| Dispel behaviour after this commit  |  |
| ------------- | ------------- |
| Alliance dispel friendly Horde MC victim  | dispels debuffs |
| Alliance dispel friendly Alliance MC caster | dispels debuffs |
| Alliance dispel friendly NPC MC victim | dispels debuffs |
| Alliance dispel hostile Alliance MC victim | dispels MC first and then random buffs |
| Alliance dispel hostile Horde MC caster | dispels random buffs including MC |
| Alliance dispel hostile NPC MC victim (friendly to Alliance before the charm) | dispels MC first and then random buffs |
| Alliance dispel hostile NPC MC victim (hostile to Alliance before the charm) | dispels random buffs |


ohganot:
Alliance dispel hostile Alliance MC victim | dispels MC first and then random buffs
-- | --

Does this mean that using Dispel Magic will dispel MC + 1 random buff? It should only dispel MC.

GuybrushGit:
Would need concrete evidence of that. I saw posts saying 2 aura dispels will remove MC + a buff.

ohganot:
Take a closer look at the videos posted in the bugreport. Both in the Lucifron video and the Orgrimmar video you can see that only the MC is getting dispelled.

Besides, Dispel Magic clearly states: "Dispels magic on the target, removing 2 harmful spells from a friend or 2 beneficial spells from an enemy."

It can do either one or the another. It can't dispel harmful and beneficial spells at the same time.

GuybrushGit:
It makes sense but we can't tell which rank dispel magic they used.
This thread has multiple people saying it removes MC + 1 buff: http://www.mmo-champion.com/threads/647217-Dispel-Magic-on-mc-ed-targets

ohganot:
Fair enough. As long as it can be countered by using Rank 1 Dispel Magic, which removes only 1 buff or debuff, its fine.

Also, if a Mind Control effect is dispellable, it should have Dispel Type: Magic associated with it. If not, it's n/a. 

http://db.vanillagaming.org/?spell=785 Prophet Skeram Mind Control isn't dispellable.
http://db.vanillagaming.org/?spell=22667 Nefarian Mind Control isn't dispellable.
http://db.vanillagaming.org/?spell=20740 Lucifron Mind Control is dispellable.

GuybrushGit:
That's not an issue, non-dispellables with continue not dispellable.

Just added a fix to Dispel / Devour Magic never resisting when removing buffs.

GuybrushGit:
A previous fix to farsight spells was causing MC to never resist. 
Reverted it and fixed farsight properly: now it can resist but won't agro when it does.

Issue #1446, opened 2017-07-12
Major bubble hearth bug - lose 10% durability superman flying bug


Mostprovidedlinks:
I hearthed from tarren mill
https://www.youtube.com/watch?v=5F8XAOV-X9I

StadenElysium:
Are you able to reproduce this?

Mostprovidedlinks:
Sure

StadenElysium:
Are you able to post instructions on how to reproduce this? :thinking:

Mostprovidedlinks:
1. I was at tarren mill and heartged from attacking the tarren mill flight master.

2. I teleported to stormwind

3. I died and lost durabity and magically flew 200km/h (125mph) in a forward direction unable to control my body

4. The whispers or chat of people still remained there 

5. I logged back on and my corpse was in stormwind

Mostprovidedlinks:
Im hoping this was fixed.

PR #2326, opened 2017-08-30
Fix invalid death state when porting to graveyard

GuybrushGit:
Fixes https://github.com/elysium-project/server/issues/2315

PR #2685, opened 2017-09-21
Prevent death during teleport

GuybrushGit:
Fixes https://github.com/elysium-project/server/issues/1446

This prevents the player from dying mid teleport, causing the corpse flying across the map issue. 
Player still takes damage but can't die.


brotalnia:
I think the issue happens only with long distance teleports where you go through a loading screen. At least that was the case with a similar issue. When you swim to the bottom of the ocean, it kills you and teleports you to the neareast graveyard at the same time, which makes the death not register in the client. So the server knows you are dead and constantly tells you "You are dead" when you try to do something, but the client thinks you are alive and you have to relog to fix it.

GuybrushGit:
It happens with near teleports as well, easy to reproduce.

Mostprovidedlinks:
yes happens with near just tried it today still happened, was only at cathdral hearthing to sw inn

GuybrushGit:
@brotalnia that invalid death state issue is fixed with https://github.com/elysium-project/server/pull/2326

thepparker:
What if we take the other route and block the teleport if the player dies? I'm assuming it occurs because the player dies between the server issuing a teleport and the client responding with a teleport ack.

GuybrushGit:
Yeah that's how the core is working right now. 
The player dies so the teleport doesn't go through, but before the player died the new location was already sent to the client, causing the corpse sliding across the world issue.

Maybe we can force another teleport when he dies during the first teleport but I have no idea how the client will respond to that, seems too convoluted of a solution and I don't know enough about the core to implement it.

Giving the small invulnerability window is simple and won't cause any issues except with blink. Can just add an exception for blink.

forpotu:
LOL HEY POTU!! I HOPE THERES NO BODY ON ELIYSUM WITH PERSON INFO ABOUT YOU OR SOCIAL MEDIA YOU BE SURPRISED HOW CRAZY SOME PEOPLE ARE IF THEY GET THAT INFO ON A FGT LIEK YOULOL HEY POTU!! I HOPE THERES NO BODY ON ELIYSUM WITH PERSON INFO ABOUT YOU OR SOCIAL MEDIA YOU BE SURPRISED HOW CRAZY SOME PEOPLE ARE IF THEY GET THAT INFO ON A FGT LIEK YOULOL HEY POTU!! I HOPE THERES NO BODY ON ELIYSUM WITH PERSON INFO ABOUT YOU OR SOCIAL MEDIA YOU BE SURPRISED HOW CRAZY SOME PEOPLE ARE IF THEY GET THAT INFO ON A FGT LIEK YOU

PR #2715, opened 2017-09-23
Missing Diplomat: Enhanced Edition.

kampeador:
This PR is a continuation of my "The Missing Diplomat" rewrite. https://github.com/elysium-project/server/pull/1395
All SD2 placeholders were rewritten, but it still was nowhere near the retail quality level. I've got some free time and returned to this quest chain again.

Missing Diplomat id: 1447:
Complete event rework featuring:
 - A new mechanics for thugs.
   * Their spawn positions have been corrected.
   * Implemented a cutscene when Dashel has been defeated and thugs stayed alive.
   * "Ok Boss!" placeholder was replaced with 2 blizzlike texts.
   * Fixed a bug with an incorrect despawn.
   * Thugs will now focus player who started a quest event and will not put other players nearby into a combat mode.
   * Thugs will now use a correct "attack" animation instead of switching to "stand" after each hit.
 - Fixed some spelling errors in existing texts.
 - Fixed a bug when Dashel sometimes get stuck in combat mode.

Missing Diplomat id: 1324:
 - A new text quote for Hendel's friends.
 - Fixed some spelling errors in existing texts.
 - Added a visual effect for the Teleport spell used by Tervosh on Private Hendel.
 - Fixed an inconsistent behavior, when a Private Hendel has been defeated, allies have not yet arrived and a random creature started to attack him.

Missing Diplomat id: 1265:
Full quest event implementation.
 - A new quest area-trigger. 1667 instead of 330.
 - Tervosh will now appear, once player enters a Sentry Point tower. All guards inside will salute him.
 - Tervosh will will buff players with Proudmoore's Defense on quest complete.
 - Many rare cases for high population has been resolved(i.e many players are doing this quest at the same time).

Missing Diplomat id:1249:
Complete event rework featuring:
 - Detail emotes for quest text.
 - Tapoke "Slim" Jahn now uses a correct stealth spell with 8 sec duration.
 - A waypoint path for Tapoke "Slim" Jahn has been reworked completely.
 - Tapoke "Slim" Jahn has learned a new ability: Pummel.
 - Implemented a Mikhail's gossip script. His gossip behavior is very different from other NPCs in a game and consists of 3 states, but only 1 is active at a time, depending on a player's quest status.
   * Gossip: A special gossip text for players trying to start a quest event if Tapoke Slim Jahn is unavailable(event was started, he is dead or pulled away from the inn).
   * Quest: Other Missing diplomat quests or event q:1249 is ready to start.
   * Nothing: No quests, no gossip.
 - Implemented a Slim's friend behavior. He will continue to help Tapoke "Slim" Jahn after enemy Vanish, Feign Death, etc.
 - Fixed a critical issue with respawn timer. It was 2 seconds after the first player completed this event.
 - Quest will also fail if Tapoke "Slim" Jahn ran away from player/group range.
 - Fixed an inconsistent behavior when Tapoke "Slim" Jahn was attacked by player from the opposite faction, during the event.

Misc:
 - Fixed an issue when: if (uiDamage > m_creature->GetHealth() || m_creature->GetHealthPercent() < 20.0f) fails and NPC dies if multiple harmful spells land on the target at the same time.

npc_escortAI API:
 - Added a method to turn on/off pathfinding between waypoints.
 - Added a method to control a delay before the first waypoint. Hardcoded value (2500ms) is too high for some quests.


forpotu:
LOL HEY POTU!! I HOPE THERES NO BODY ON ELIYSUM WITH PERSON INFO ABOUT YOU OR SOCIAL MEDIA YOU BE SURPRISED HOW CRAZY SOME PEOPLE ARE IF THEY GET THAT INFO ON A FGT LIEK YOULOL HEY POTU!! I HOPE THERES NO BODY ON ELIYSUM WITH PERSON INFO ABOUT YOU OR SOCIAL MEDIA YOU BE SURPRISED HOW CRAZY SOME PEOPLE ARE IF THEY GET THAT INFO ON A FGT LIEK YOULOL HEY POTU!! I HOPE THERES NO BODY ON ELIYSUM WITH PERSON INFO ABOUT YOU OR SOCIAL MEDIA YOU BE SURPRISED HOW CRAZY SOME PEOPLE ARE IF THEY GET THAT INFO ON A FGT LIEK YOULOL HEY POTU!! I HOPE THERES NO BODY ON ELIYSUM WITH PERSON INFO ABOUT YOU OR SOCIAL MEDIA YOU BE SURPRISED HOW CRAZY SOME PEOPLE ARE IF THEY GET THAT INFO ON A FGT LIEK YOULOL HEY POTU!! I HOPE THERES NO BODY ON ELIYSUM WITH PERSON INFO ABOUT YOU OR SOCIAL MEDIA YOU BE SURPRISED HOW CRAZY SOME PEOPLE ARE IF THEY GET THAT INFO ON A FGT LIEK YOU

src/scripts/eastern_kingdoms/stormwind_city/stormwind_city.cpp
```diff
@@ -104,313 +104,352 @@ CreatureAI* GetAI_npc_bartleby(Creature* pCreature)
...
+                    m_nextPhaseDelayTimer = 3000;
+                    m_eventPhase = MDQP_SAY2;
+                }
+                else
+                    m_nextPhaseDelayTimer -= uiDiff;
+            } break;
+        case MDQP_SAY2: // Occurs only if thugs are alive
+            {
+                if (m_nextPhaseDelayTimer < uiDiff)
                 {
```
thepparker: You should use helper methods for blocks like this, so you don't have to duplicate the same code for each minion. There are a few blocks here that can be reduced to some simple methods

src/scripts/eastern_kingdoms/stormwind_city/stormwind_city.cpp
```diff
@@ -104,313 +104,352 @@ CreatureAI* GetAI_npc_bartleby(Creature* pCreature)
...
-            // Set time delay same as other thugs to match DB. This will prevent too fast respawns.
-            dashelStonefistAI->m_thugs[i]->SetRespawnDelay(270); // 270 secs
-                                                                 // make them focus player
-            if (pPlayer)
-                dashelStonefistAI->m_thugs[i]->AI()->AttackStart(pPlayer);
+            // On official in some cases: he didn't say this phrase. I am not sure if it was a bug or a random feature.
+            // Dashel says: Now you're gonna get it good, "PlayerName".
+            DoScriptText(SAY_PROGRESS_1_DAS, pCreature, pPlayer);
+
+            // set dashel stonefist's faction to neutral
```
thepparker: Don't need all these comments, the code is already pretty self explanatory. Making it too verbose actually makes it harder to read

src/scripts/eastern_kingdoms/stormwind_city/stormwind_city.cpp
thepparker: You should use helper methods for blocks like this, so you don't have to duplicate the same code for each minion. There are a few blocks here that can be reduced to some simple methods

src/scripts/eastern_kingdoms/stormwind_city/stormwind_city.cpp
```diff
@@ -104,313 +104,352 @@ CreatureAI* GetAI_npc_bartleby(Creature* pCreature)
...
+            pCreature->setFaction(FACTION_NEUTRAL);
+            // prevents random player pick up a quest during event
+            pCreature->RemoveFlag(UNIT_NPC_FLAGS, UNIT_NPC_FLAG_QUESTGIVER);
+
+            // save player guid for later use(to trigger a quest completed)
+            dashelStonefistAI->m_playerGuid = pPlayer->GetObjectGuid();
+
+            // spawn thugs and make them focus player.
+            // thug 1
+            dashelStonefistAI->m_thugs[0] = pCreature->SummonCreature(NPC_OLD_TOWN_THUG, -8676.075195f, 443.744019f, 99.632210f, 3.981758f, TEMPSUMMON_DEAD_DESPAWN);
```
thepparker: Need to check if the summon succeeded before attempting to use the pointer it would normally return when successful

src/scripts/eastern_kingdoms/stormwind_city/stormwind_city.cpp
kampeador: You are right. There are too much comments in my code. The reason for this is that most of them were written before i've started working on the script. Some of them were used to comment a packets from the sniff dump, which is not self explanatory. I'll remove some them.

src/scripts/eastern_kingdoms/stormwind_city/stormwind_city.cpp
kampeador: Do you mean a DialogueHelper class or just move to duplicates to the separate method?
1. DialogHelper has a limitation. Thugs have the same "entry".
2. I thought about this, but there are only 2 of them. Each thug has its own text, timer and waypoint. Code is "linear". It is easy to read and debug each "phase". It is also easy to change in case thug behavior will differ from each other.
A problem is that this cut-scene was buggy on retail in some versions. The whole Missing Diplomat quest chain was buggy from time to time. I've reported a lot of bugs on retail. There can be additional texts for thugs. Same goes for Hendel's friends in Dustwallow. Currently texts are hardcoded for them, but i'm pretty sure they were differ. I've implemented only the ones i saw. Obviously, 3 times was not enough to complete this quest chain to gather 100% data. Currently i can't provide it. Maybe in future someone(or me) can easily add that missing data and according to that, some code can be cleaned/optimized.

src/scripts/eastern_kingdoms/stormwind_city/stormwind_city.cpp
kampeador: I've tried to trigger server crash on every pointer check.
Its a good practice to check everything. I can add them, but in real scenario these checks don't give any safety.
It can fail in 2 cases:
1. Map local guid overflow.
https://github.com/elysium-project/server/blob/development/src/game/Maps/Map.cpp#L4271
Currently, core can't reload map to prevent this, so it will lead to the server shutdown anyway.

2. Corrupted DB or missing data. Core will crash in 99% cases before it reaches my script.

I am pretty sure there are a lot of gaps here.
For example:
https://github.com/elysium-project/server/blob/development/src/scripts/kalimdor/tanaris/zulfarrak/zulfarrak.cpp#L695-L698

src/scripts/eastern_kingdoms/stormwind_city/stormwind_city.cpp
thepparker: Just create a method that does the functions which are similar for each summon, and store the data that changes between the two in a struct array (i.e. positions). Pass the script text ID as a param. There's no need to have them separately in case behaviour is different in the future, it's unlikely any further changes would be made to the event, and we know it doesn't change in terms of actual gameplay or design

PR #2725, opened 2017-09-24
Remove fish from wetlands lake

GuybrushGit:


src/game/Objects/GameObject.cpp
```diff
@@ -736,6 +736,10 @@ void GameObject::getFishLoot(Loot *fishloot, Player* loot_owner)
     uint32 zone, subzone;
     GetZoneAndAreaId(zone, subzone);
 
+    // Don't allow fishing in hidden wetlands lake
+    if (subzone == 11 && loot_owner->IsWithinDist2d(-4074.74, -1315.79, 100.0f))
```
Kittnz: This is a huge hack

GuybrushGit: Yeah I know, but how else are you gonna remove loot from a specific lake with no subzone?

PR #2741, opened 2017-09-24
Make sure swing timer is reset at end of cast.

brotalnia:
The problem was that Spell::finish(), where the reset timer is reset, was called AFTER the unit was already made able to attack, so resetting it there didn't accomplish anything in some situations. Now auto attack will no longer go off immediately after finishing a cast.

### DIFF
--- a/src/game/Spells/Spell.cpp
+++ b/src/game/Spells/Spell.cpp
@@ -3494,6 +3494,13 @@ void Spell::cast(bool skipCheck)
         }
     }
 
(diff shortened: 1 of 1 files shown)


Cabro:
Works the same for both players and NPCs?

brotalnia:
It should.

BrynjaDarrow:
Uh, this might be a really important fix1

Marcelo20XX:
OK I tested it but mobs instant spells dont reset the attack timer so can you add the fix for this too? thanks

brotalnia:
I don't think most instant spells should do that?

Issue #2482, opened 2017-09-09
Hunter Serpent String resists (wrong mechanic)


Crusaderz:
So i noticed my serpent shot resists sometimes in pve (and most of the time in pvp against red or skull level targets). I played retail vanilla (and later) and never had serpent shot resists. SS is not spell and it always should apply to the target. 

Here's some proofs:

https://www.youtube.com/watch?v=c5qmC0zKh3Y 25 sept 2006 Lvl 49 vs Red dragonspawn (56-58). SS always lands.
https://www.youtube.com/watch?v=WG9UB90DZz8 oct 23 2006, level 28 kites level 35 Stitches. SS always lands.
https://www.youtube.com/watch?v=jt5BepzK7oM jul 6 2006, level 39 vs 60 rogue. SS always lands.
https://www.youtube.com/watch?v=2PjMESVAWL0 Level 32 kites Skeletal Flayer in Sorrow Hills (50-51) 26 aug 2006. SS always lands.
https://www.youtube.com/watch?v=SLpHjG24k-c Level 15 kills level 40+ (mounted) afk Paladin. 12 nov 2006, SS always lands.
https://www.youtube.com/watch?v=V8vLQiLfDV0 uploaded in feb 2008, most likely BC but the point is - SS has 100% chance to land in any wow expansion.
http://www.arenajunkies.com/topic/21094-serpent-sting-through-cos/ 2008, rogue complains about SS through cloak of shadow, first replay: "Yes, since it isn't really a spell. Pretty sure i've never had anyone resist a Serpent Sting in 3 years." This adds to the evidence.

Out of 6 videos and many SS shots none of them fails against red-skull targets so current SS version on Elysium is cearly bugged.

Sorry for my english.

PR #2746, opened 2017-09-24
Fix to Serpent Sting and Arcane Shot

GuybrushGit:
Fixes https://github.com/elysium-project/server/issues/2482

The Serpent Sting spell_mod to make it scale with spelldamage had the side effect of making it use spell resist calculations.
Removing it fixes it, its not necessary anymore since I already fixed all ranged hits not scaling in a previous commit.

Also removed the Arcane Shot spell_mod that made it never miss, no idea why that was there, there are multiple mentions of Arcane Shot full resists showing up as a miss in the combat log.


Crusaderz:
Super. 
Can you please also check imp. Concussive Shot? I'm pretty sure proc counts as a spell right now, since it's **always** resists against red-skull level players, mobs. Not kidding. Always. The proc effect should be physical: 
talent - http://db.vanillagaming.org/?spell=19407 
proc - http://db.vanillagaming.org/?spell=19410 
Proc indeed has small resist chance (added in 1.7):
http://wowwiki.wikia.com/wiki/Patch_1.7.0

> Improved Concussive Shot - The stun effect can now be resisted.

But it's just chance to fail and sadly the correct resist chance in unknown.
Thanks.

GuybrushGit:
Imp Concussive Shot is resist as a spell, as it should.

Crusaderz:
> Imp Concussive Shot is resisting as a spell, as it should.

Found video (2006) of "skull" mob kiting, where stun landed only once out of ~6 procs, so it's true.
Thanks!

Issue #2742, opened 2017-09-24
Spells that are shield/absorb type have 0 threat, when they should have threat.


Cabro:
For example priest's shield has 0 threat from what i noticed, when it should be a percentage of the amount absorbed.

GuybrushGit:
>Tests have concluded that Power Word: Shield causes approximately half as much threat as healing for the same amount. So threat-wise Power Word: Shielding for 1000 is roughly equivalent to healing for 500. However, the spell does not cause any threat, if cast before you or your target are in combat.
>https://wow.gamepedia.com/index.php?title=Power_Word:_Shield&oldid=317147

Cabro:
Yes, i wasnt fully clear. I didnt say you should get threat when the damage you take is absorbed, but how much the shield absorbs in the moment of the cast.
My bad for that. Is it working like this on the server? It doesnt seem to be giving threat to the caster, or i could just be wrong.

Cabro:
I see the fix, happy to see it being worked on.
However, it's all spells that have the same functionality (absorb trinkets, T3 procs etc.).
I ask you kindly to fix them all!

GuybrushGit:
Give sources then. Threat addons have threat for PWS but not the others.

Cabro:
It seems you are right, saw the Ice Barrier tag with "generate no threat", seems only PW:S is the one that gives threat on cast, indeed.
+1

GuybrushGit:
Not having the no threat flag is not enough to change it, there are plenty of no threat spells that don't have it.  
There's no mention any target_self absorb generating threat. PWS is the only absorb that can be cast on anyone.

Cabro:
Ok then!

PR #2764, opened 2017-09-25
Implemented Power Word: Shield threat

GuybrushGit:
Fixes https://github.com/elysium-project/server/issues/2742

Side note: Vanilla KTM doesn't have PWS threat but vanilla wowwiki mentions it having threat. TBC Omen2 has threat on PWS.


ItsmeAga:
shouldnt that be the case for all absorbs? not just one single spell?

GuybrushGit:
I found no evidence for other absorbs and all the others are self cast only.

Maievh:
Whoever added that to the vanilla wiki is wrong. Shield has never generated threat.

PR #2767, opened 2017-09-25
Increased Jazzrik Bloodsail reputation

GuybrushGit:
Custom change suggested here: https://forum.elysium-project.org/topic/55229-make-jazzrik-give-5-bloodsail-reputation-early/


brotalnia:
But why? We doing "make a wish" custom changes now? And even if there was a staff decision to do this, it should not be added as migration, thus contaminating the db with incorrect data. Custom changes should be in separate files that can be optionally applied after setting up the realm.

Nelythia:
Hey, Nelythia (the guy suggesting this) here.

As written in my suggestion linked by Guybrush the required time to grind this in Vanilla is around 75 days of raw /played time.
This NPC is the only NPC ingame until Cataclysm that does give rep after reaching revered status with the Bloodsail Buccaneers

Unfortunately this NPC has a 5 minute respawn timer, which gives you at best 12 rep an hour.
This grind was done by players on actual retail, however they were not as experienced back then and were stuck on 20999/21000 due to doing the 3 quests previously, instead of saving them until you hit revered cap.

While it is possible to do this grind even without any change, the time required to actually do this is absurd and honestly in retail vanilla/tbc there weren't as many players commited to this kind of gameplay.
It was over time way easier grinding him even for +1 rep and achieving it eventually, however as of 2017 and knowing everything about every expansion and having players trying to accomplish literally everything within this game, I believe an increasement of reputation is justified.

Maybe it should not be +25 as suggested by Rakesh in the thread, but if its +5 or MAYBE even +10, I don't see any issue and it should be justified to fit the population.
Just like dynamic respawn on NPCS/Quest objects/Herbs+Ores in general as well.

Of course, that is your decision and I respect and accept any decision you and the staff will take.
I, and many other rep grinders, would however sincerely appreciate if you could give it some thought, although we do understand its by far not a high priority on your list.

Phoosy:
Making custom changes because of what amounts to a meme-request at best seems counterproductive to the intent of the Project.

thepparker:
This is a pretty frivolous change that cheapens the grind to exalted in my opinion. If you wish to go for the title of Bloodsail Admiral, knuckle down and enjoy the grind.

Issue #2743, opened 2017-09-24
mindcontol +blessing of protection bug


showabug:
Currently, in warsong gulch you can use mindcontrol + blessing of protection on enemy flag carrier to force him to drop the flag. This isn't working as intended as blizzard fixed this bug in patch 1.7:
Offensive use of damage immunities no longer causes the flag to drop in Warsong Gulch.
http://wowwiki.wikia.com/wiki/Patch_1.7.0

TimEsfandiari:
#RetPrio

PR #2769, opened 2017-09-26
Fix to BoP flag drop while charmed

GuybrushGit:
Fixes https://github.com/elysium-project/server/issues/2743


src/game/Spells/SpellAuras.cpp
```diff
@@ -3678,8 +3678,9 @@ void Aura::HandleAuraModSchoolImmunity(bool apply, bool Real)
     target->ApplySpellImmune(GetId(), IMMUNITY_SCHOOL, m_modifier.m_miscvalue, apply);
 
     // remove all flag auras (they are positive, but they must be removed when you are immune)
-    if (GetSpellProto()->AttributesEx & SPELL_ATTR_EX_DISPEL_AURAS_ON_IMMUNITY
-            && GetSpellProto()->AttributesEx2 & SPELL_ATTR_EX2_DAMAGE_REDUCED_SHIELD)
+    if (apply && GetSpellProto()->AttributesEx & SPELL_ATTR_EX_DISPEL_AURAS_ON_IMMUNITY
+              && GetSpellProto()->AttributesEx2 & SPELL_ATTR_EX2_DAMAGE_REDUCED_SHIELD
+              && target->IsPlayer() && !target->isCharmed())
```
thepparker: I'm not sure if any creatures use this behaviour, but it's probably worth allowing it to still be called on non-player units as well,

GuybrushGit: I checked, all 7 spells with this flag are cast on players only:
"21546"
"23333"
"23335"
"23763"
"23764"
"24314"
"29519"

thepparker: Ok, great

Issue #2094, opened 2017-08-14
Darrowshire - Cenarion Hold NPC's does not activate


Ripper007:
The NPC;'s in Cenarion Hold ( Infantry, etc) should activate and attack anyone in the hold if players attack other NPC's ( for ex. the opposite faction flight master). That is not happening at this moment. One alliance mage attacked, kited and killed the horde flight master without impunity. That should no be possible.

PR #588, opened 2017-05-02
Fixes to the PvP flag mechanics

GuybrushGit:
This fixes the following:

- Direct and indirect AoE from non flagged players damaging flagged players
- Direct AoE on flagged players turning the caster's pvp flag on
- Right clicking flagged players outside melee range turning pvp flag on
- Attacking or healing PvP flagged NPCs not flagging the caster
- Flying through contested areas flagging you for pvp - now only flags if you land in the contested area
- PvP quests not flagging the player
- Getting flagged when we challenge a flagged player to a duel
- Pets attacking a flagged unit not flagging the pet owner

Issues & sources:

https://github.com/elysium-project/server/issues/365
https://elysium-project.org/bugtracker/issue/1795
https://elysium-project.org/bugtracker/issue/1746
https://elysium-project.org/bugtracker/issue/1405
https://elysium-project.org/bugtracker/issue/1045
https://elysium-project.org/bugtracker/issue/976
https://elysium-project.org/bugtracker/issue/962
https://elysium-project.org/bugtracker/issue/658

PR #2959, opened 2017-10-02
Fixes to the guard AI

GuybrushGit:
- Fixed attacking PvP flagged mobs not flagging the player for contested PvP (https://github.com/elysium-project/server/issues/2094).
- Fixed guards not protecting nearby friendly NPCs.

### DIFF
--- a/src/game/AI/ScriptedGuardAI.cpp
+++ b/src/game/AI/ScriptedGuardAI.cpp
@@ -191,6 +191,50 @@ void guardAI::UpdateAI(const uint32 diff)
     }
 }
 
+void guardAI::MoveInLineOfSight(Unit* pWho)
+{
+    if (!m_creature->IsWithinDistInMap(pWho, m_creature->GetAttackDistance(pWho)))
+        return;
+
+    if (m_creature->CanInitiateAttack() && pWho->isTargetableForAttack())
+    {
+        // Attack hostile targets
+        if (m_creature->IsHostileTo(pWho) && pWho->isInAccessablePlaceFor(m_creature) && m_creature->IsWithinLOSInMap(pWho))
+        {
+            if (!m_creature->getVictim())
+                AttackStart(pWho);
+            else if (m_creature->GetMap()->IsDungeon())
+            {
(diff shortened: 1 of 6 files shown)


BrynjaDarrow:
Wasn't the first part already fixed with https://github.com/elysium-project/server/pull/588?
"Attacking or healing PvP flagged NPCs not flagging the caster"

GuybrushGit:
@BrynjaDarrow no its a different flag.

GuybrushGit:
Removed the AI changes they did not work if the player never moved.
Replaced with a faction check that does the same without adding a bunch of code to the AI. 
There is no flag or identifier for contested factions so I had to hardcode the ids.

brotalnia:
Actually you can check the faction template flags instead of hardcoding the faction ids.

FACTION_TEMPLATE_FLAG_CONTESTED_GUARD = 0x00001000 (4096)

Here is a list of faction templates with it:
> Template 121, Faction 21, Flags 6177, FriendlyMask 0, HostileMask 8
> Template 475, Faction 369, Flags 6177, FriendlyMask 0, HostileMask 8
> Template 637, Faction 470, Flags 6177, FriendlyMask 0, HostileMask 8
> Template 854, Faction 577, Flags 6177, FriendlyMask 0, HostileMask 8
> Template 1254, Faction 609, Flags 4129, FriendlyMask 0, HostileMask 8
> Template 1615, Faction 169, Flags 6177, FriendlyMask 0, HostileMask 8
> Template 1625, Faction 529, Flags 4129, FriendlyMask 0, HostileMask 8
> Template 1635, Faction 169, Flags 6161, FriendlyMask 0, HostileMask 0

As you can see those are all the factions you listed.

GuybrushGit:
Those are for the guards, we want the units protected by the guards. For example for BB its 120 not 121.

GuybrushGit:
But then I'd need to iterate through all the faction templates just to match factions that equal the faction of a contested guard. 
This function gets called every time someone deals damage to a unit, it seems a waste of CPU to do that when the factions are constant.

Issue #1588, opened 2017-07-20
NPC and MOBs in Silithus stay in a  state of strike for days since 1.10 patch on Anathema


veraandtaissa:
Encrypted Twilight Text and Twilight Cultist set do not have any drop-rate .

NPC named Vargus who should give players Volunteer's Battlegear ,Veteran's Battlegear Champion's Battlegear and Stalwart's Battlegear is still striking...

so far so these ....ty

veraandtaissa:
and still bugged at 2017/7/21

veraandtaissa:
and still bugged on 2017/7/22 ，just marked for check ,im not Urging

Lufhtrae:
I am able to confirm that you are unable to turn the quest in even if you already had it in your quest log, and you are unable to pick up the quest after abandoning it. This issue started about two days ago which was posted on the forums initially. GM in-game confirmed that this is a known issue and is being worked on.
Thread - https://forum.elysium-project.org/topic/50606-stalwarts-battlegear-quest/

cloudx123:
bumb..hope they fix soon got 11 people in my guild who cant complete it

PR #2972, opened 2017-10-04
Fix to event start dates

GuybrushGit:
Some event start dates were set at 2025 to prevent them from starting. 
No longer necessary with the progressive database controlling when they start.

Possibly the issue behind https://github.com/elysium-project/server/issues/1588


thepparker:
Very good point, these probably should have just been disabled while keeping reasonable start/end times in the original DB (aka the same as the Patch events)

tacodoctor:
this is still an ongoing issue on the anathema server

Issue #2242, opened 2017-08-24
Blasted Land Buffs should give +50 stats until 1.11


BrynjaDarrow:
Blasted Land buffs "should give +50 to their respective stats instead of +25 until 1.11:

https://web.archive.org/web/20060407225314/http://www.thottbot.com:80/?i=6228
https://web.archive.org/web/20060222234625/http://www.thottbot.com:80/?i=5175
https://web.archive.org/web/20060314032014/http://thottbot.com:80/?i=6261
etc."

quote by fgsfdss in #2197

GuybrushGit:
Will not be implementing this, see https://github.com/elysium-project/server/pull/3005

Issue #2197, opened 2017-08-21
Spirit of Zanza and Blasted Lands buffs don't stack


ohganot:
Only the +25 Stamina buff from Blasted Lands shouldn't stack with Spirit of Zanza. All others should.
I couldn't find video evidence on this, but its very clear that they should stack due to a number of reasons.

1) Buffs with different stats always stack. For example Elixir of the Mongoose and Elixir of Fortitude. It makes no sense that +50 stamina and +25 strength don't stack.

2) Buffs with big differences in durations always stack. Zanza lasts 2h, Blasted Lands buffs last 1h. For example DM +200 AP buff and Winterfall Firewater or Mighty Rage Potion and Elixir of Giants.

3) They are located in different zones and zone specific buffs generally stack with each other. For example Songflower, Rallying Cry, Spirit of Zandalar, DM Buffs, Jujus, etc.

4) There was no evidence that they shouldn't stack #275 "This was updated a while ago but if there's any evidence that it's incorrect (quite possibly is), that would be appreciated" - StadenElysium. Therefore we should go with logic and follow general buff stacking rules. A buff that gives stamina should stack with a buff that gives strength.

Please help me find evidence for this or simply fix it to be in union with the rest of buff stacking rules!

fgsfdss:
On this subject, based on a comment on thottbot Lung Juice Cocktail should give +50 stam instead of +25 until 1.11: https://web.archive.org/web/20060407225314/http://www.thottbot.com:80/?i=6228

Comment confirming patch change: https://web.archive.org/web/20070303204527/http://www.thottbot.com:80/?i=6228

fgsfdss:
On this subject, these items should give +50 to their respective stats instead of +25 until 1.11: https://web.archive.org/web/20060407225314/http://www.thottbot.com:80/?i=6228
https://web.archive.org/web/20060222234625/http://www.thottbot.com:80/?i=5175
https://web.archive.org/web/20060314032014/http://thottbot.com:80/?i=6261
etc

blubba22:
I would open a new issue with the wrong stats since these are two different issues. 
I sure do hope they fix both.

GuybrushGit:
>You can use them with a diffrent BL buff. 
>I.e. I stack the health one with the agi BL buff for rag
>https://web.archive.org/web/20070314082028/http://www.thottbot.com:80/i20079

PR #3005, opened 2017-10-05
 Blasted Lands buffs stats and stacking rules.

GuybrushGit:
- https://github.com/elysium-project/server/issues/2242 Implemented the +50 stats pre-1.11, had to hardcode it since Blizzard just modified the spell. The client will still display 25 but players will get 50.

 - https://github.com/elysium-project/server/issues/2197 Fixed the stacking rules. New rules are: only 1 zanza buff allowed, only 1 BL buff allowed and only 1 between Spirit of Zanza / BL Stamina / BL Spirit allowed.


brotalnia:
I thought we aren't doing spell progression cause the spells look wrong in the client? If we want that, i could make it so spells are loaded from the db and it will load the table for the appropriate patch, giving you full patch by patch spell progression with all the old effects, base points and other changes listed in the patch notes, without needing to hardcore things like this.

GuybrushGit:
Yeah its debatable, its a spell change but its still related to itemization. Should I remove that part and just leave the stacking fix?

brotalnia:
Yes i feel it's best to take it out. If we want it changed on live realms, it should be done in spell_effect_mod instead of hardcoding the basepoints in the core. There's a custom sql folder now -> https://github.com/elysium-project/server/tree/development/sql/custom

thepparker:
Leave the stats as they are in 1.12, stacking looks okay.

GuybrushGit:
Reverted

PR #3014, opened 2017-10-05
Grouped players (not in dungeons) will now share rep gains instead of…

NolinNostralia:
… receiving the full amount


NolinNostralia:
thanks to @KelNostralia

druidish:
Reputation from mob kills shouldn't be split between players.... Everyone should receive the full amount.

PR #3088, opened 2017-10-10
Implement restoration of deleted items.

brotalnia:
When a player has an item that does not exist, it will now be saved in a new table called `character_deleted_items` so that those items may be restored at a later patch. The table is checked on server startup and items that exist in the current patch get sent to the player by mail.

restored items

Now you can sleep sound knowing your contraband items that you should have never obtained in the first place are safe.


Mostprovidedlinks:
@brotalnia will this be made so people on anathema can transfer their chars to elysium and the items that are for a later patch are kept for later?

brotalnia:
This is a nice bonus use of it that i didn't think of when making it, but yes it would be useful in such a situation.

PR #3112, opened 2017-10-11
Added an option to modify timers inside movement generators.

kampeador:
A new method: setMoveGenDelayRequest has been added. Which allows to safely modify(if possible) timer inside a movement generator, which allows a blizzlike interaction between Player and Creature.

For example: Player interacts with a traveling merchant or a quest giver and the NPC accidently starts moving without waiting a player. In some cases, its impossible to read quest texts and other things, because a dialog will close when the NPC is too far away.
Almost every private server has this issue. Current implementation is using a hardcoded STOP_TIME_FOR_PLAYER = 30 seconds, inside waypoint movement generator. RandomMovementGenerator has no waiting timer at all, DB uses a global hack to disable it for quest givers, vendors etc, but some of them are still using it.

The following features had been implemented:
 - Reverted commit: cb855707f6109b11c7fae3978b438a8ebd325630. It is not blizzlike, it produces a lot of bugs with creature's orientation in gossip scripts with a gossip option. All players around will now see the same orientation and it will not be accidentally modified anymore(check out my Argent Medic implementation as an example).
 - A waiting timer for the RandomMovementGenerator. DB hacks are not needed anymore.
 - NPC will wait for 3 minutes after player interaction.
 - If an NPC is already in waiting state and waits on waypoint or previously interacted with the player and its timer is less than 3 minutes, it becomes 3 minutes, otherwise, timer will not be modified.
 - Waiting timer affects NPCs which are quest givers, vendors, battleground masters or use gossips.
 - No pointer checks or unsafe casting. Each MoveGen is unique and method's behavior is implementation dependant.
 - Timer is not hardcoded and can be used inside scripts, to stop an NPC for some time, without any additional timers and movement generators.(Used in my latest version of "The Missing Diplomat q:1265" and other quests, which are not published yet)
 * Creature groups/formations are not affected. Only a leader will use a waiting timer and all his group will stop. Formation system is a separate topic and it has some limitations which prevent a blizzlike behavior(leader shouldn't be static, etc), so i haven't touched them here.

### DIFF
--- a/src/game/Handlers/BattleGroundHandler.cpp
+++ b/src/game/Handlers/BattleGroundHandler.cpp
@@ -55,6 +55,9 @@ void WorldSession::HandleBattlemasterHelloOpcode(WorldPacket & recv_data)
     if (!pCreature->IsStopped())
         pCreature->StopMoving();
 
+    // Request creature to wait for some time if possible.
+    pCreature->setMoveGenDelayRequest(STOP_TIME_FOR_PLAYER);
+
     BattleGroundTypeId bgTypeId = sBattleGroundMgr.GetBattleMasterBG(pCreature->GetEntry());
 
     if (bgTypeId == BATTLEGROUND_TYPE_NONE)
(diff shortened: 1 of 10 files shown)


ImperfectCell:
Great job

ryancheung:
Why closed?

Issue #3083, opened 2017-10-10
Death Talon Hatchers interruptable


GuybrushGit:
Hatchers right now are not immune to Counterspell, is there any reason why?

I found 0 mentions on guides to Hatchers being interruptable and retail videos never show them being interrupted as well:
http://leftoversraiding.org/index.php/topic/772/1/blackwing-lair-summary---pt.-1-(intro-razorgore-vaelastrasz-broodlord).html
http://wowwiki.wikia.com/wiki/Broodlord_Lashlayer#The_Suppression_Rooms
https://www.youtube.com/watch?v=-C4-1MPEPF0
https://www.youtube.com/watch?v=EtRvIDX2eMg

Sabre142:
Needs more research.

Imbaslap:
it makes 0 sense why a mob would be susceptible to only Counterspell, but not kick/pummel/shield bash.

the current state of the Hatchers are basically 1 in a million throughout multiple expansions. this is most likely due to poor research done by open source, which then got passed on from server to server.

https://github.com/ccshiro/cc-buglist/issues/2500
i think the issue falls on projects accepting claims for how things were without any sort of evidence backing up the claim.

Issue #3082, opened 2017-10-10
Razorgore Cleave range


GuybrushGit:
Currently Razorgore is doing his 10 man cleave, but the range is so short (5yd) that it has no effect. On every retail video either the melee don't join the fight or they all get hit by the cleave.
Every 10 man cleave in the game is 5yd so Razorgore's cleave range has to be modified, similar to what was done for Vael.

At 7:40 every melee is behind the boss and they still get hit by the cleave:
https://youtu.be/8YZsjXJsxxs?t=455

ItsmeAga:
probably he had a 360° cleave on retail, like a few other bosses

Imbaslap:
it should chain just like Vael's Cleave.

Issue #2955, opened 2017-10-02
aggro ranges in blackwing lair


bgfds:
https://www.twitch.tv/videos/179189773?t=24m02s
The aggro range on these trash packs seems outrageously low. Surely these mobs should not be skippable in this manner?

Also the trash packs after Vael can be pulled by a hunter and then dragged down to Razorgores room and the raid can skip that entire room, surely the raid should be put into combat by the combat pulse?

This can also be done with the 3 big guys infront of Chromaggus, hunter can pull them and you can run in to his room and summon the hunter back.

I don't know if it will be possible to find out if these are blizzlike at all but even if they are blizzlike they should be removed. It seems pointless to allow people to skip the entire dungeon.

mytelefe:
So u literally skip 2-3 packs and that is "skip the entire dungeon".

bgfds:
4 Technician packs
1 Wyrmguard pack in that video alone

Combined with the potential to skip the wyrmguards infront of chromaggus and the potential to skip vaels trash room entirely then yes, that is pretty much skipping the entire dungeon

Dasharius:
AQ40 trash aggro radius has nothing to do with BWL Trash.

bgfds:
a guild has done a speedrun today where they have a hunter pull the entire bottom floor of the suppression room and kites it to the front of the instance without any member of the raid being in combat. They skipped an ENTIRE floor of the suppression room.

You can't seriously think this is working as intended?

GuybrushGit:
The Wyrmguard doesn't patrol, better view of the placement here: https://youtu.be/dPI3y0Kjcas?t=280

PR #3144, opened 2017-10-12
BWL fixes

GuybrushGit:
Some of these issues are debatable, evidence is scarce:

- Added disarm immunity to Wyrmguards and Captains. It doesn't seem right that these mobs can be nullified by disarm but I found no evidence either way. No guides mention disarming.
- https://github.com/elysium-project/server/issues/3083 Death Talon Hatcher interrupt immunity.
- https://github.com/elysium-project/server/issues/3151 Flamestrike increased initial timer and removed random targetting
- https://github.com/elysium-project/server/issues/3082 Razorgore's cleave range
- Gave Death Talon Wyrmguard his proper weapon
- https://github.com/elysium-project/server/issues/2955 Overseer pack corrected positioning
- https://github.com/elysium-project/server/issues/3202 Fix to Nefarian warrior call


ohganot:
If you change flamestrike to be not interruptible, you also need to make it only cast on the current aggro target. As you see in the videos posted in issue #3083 the flamestrikes are only cast on the tank.

thepparker:
While I agree the Hatchers should probably be immune to interrupts, we need to change the casting logic for them if we're making them immune. For example, you can see in the videos that they do not begin to cast Flamestrike for a solid 10 seconds, however on our server they will begin casting immediately on the pull. I think it's a bit unfair for the mob to get a free cast while standing in the middle of the room, potentially aggroing additional mobs with it.

I'm not sold on the disarm immunity, perhaps the Wymguards (big ones?) and Captains should be, but it seems to make sense the smaller ones can be disarmed. There are a few cases where disarming mobs is not well documented in Vanilla, for example the Alterac Valley Marshals/Warmasters and Generals.

Aklol:
I think the Razorgore cleave range was blizzlike. However, I don't think there's enough evidence to warrant the Hatcher change yet. If there's any change, it's that Flamestrike should be cast on the current target (typically tank).

GuybrushGit:
Removed Overseer from the disarm list, increased flamestrike initial timer and removed random targetting (confirmed on multiple videos).

Aklol:
If I'm reading the code correctly, the flamestrike target fix also applies to Spellbinders?

And we're holding off on the counterspell fix?

PR #3247, opened 2017-10-29
 Replacement of ACE threads with std

WobLight:
Refresh pr after fork mess. #324


Giperionn:
I will test that build very soon

